PR #1, opened 2017-12-14
Proposed README

secondsun:
This is a readme with an example flow for how to submit a proposal.

Sorry for a wide spam of reviewers; I want as many eyes as I can get.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,15 @@
+# AeroGear Proposals
+This repository contains proposals, examples, PoCs for potential new mobile features to be build by the mobile team.
+
+## How to submit a proposal
+* Fork this repository 
+* Create a new directory with your proposal
+* Include a README.md which is either your proposal or the index of the proposal
+* Send a PR to master on github.com/aerogear/proposals
+* Email aerogear-dev so we know to take a look
+
+## When to submit a proposal
+Proposals aren't for everything, and they are mostly for new or large features.  If you have a bug fix, documentation improvement, etc then a PR with the fix is much more appropriate.  If you have a minor feature such as exposing a logging parameter, code cleanups, or adding tests then you could probably get by with a PR as well, but you might want to ping the mailing list (aerogear-dev) to get a sanity check.  However if you are adding a major feature like adding a new algorithm to the sync server, migrating HTTP libraries from OKHttp to volley, etc then these are they types of things proposals are for.  As a rule of thumb if what you want can be fit into a single issue then you should probably not write a proposal.
+
+## What does a good proposal look like
+When you submit a proposal it should be fairly obvious what you are trying to build.  In general you should describe what you are trying to build, its core features, any points of discussion to include, etc.  If you think something will make the proposal easier to follow or clear up any misconceptions then add that as well.  For SDKs if you have some psuedo code or POC implementation, include that in your directory.  If you want to create UI mockups then feel free to include those as well.
--- a/README.md
+++ b/README.md
@@ -3,13 +3,9 @@ This repository contains proposals, examples, PoCs for potential new mobile feat
 
 ## How to submit a proposal
 * Fork this repository 
-* Create a new directory with your proposal.
-Use structure from [template](./template.md)
-* Include a README.md which is either your proposal or the index of the proposal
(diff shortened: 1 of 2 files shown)


matzew (approved):
lgtm

README.md
```diff
@@ -0,0 +1,9 @@
+# AeroGear Proposals
+This repository contains proposals, examples, PoCs for potential new mobile features to be build by the mobile team.
+
+## How to submit a proposal
+* Fork this repository 
+* Create a new directory with your proposal
```
maleck13: I am not getting the new dir for each proposal this just seems like noise. What is the benefit? Personally I would prefer somthing like
```
/cli
  command_structure.md
  goals_and_principles.md
/sdks
...
/services
...
```

README.md
```diff
@@ -0,0 +1,9 @@
...
+* Include a README.md which is either your proposal or the index of the proposal
+* Send a PR to master on github.com/aerogear/proposals
+* Email aerogear-dev so we know to take a look
```
wei-lee: From Craig's email, I suggest we also add some guidelines here about when a proposal is required and what's the basic structure should be like.

* When a proposal is necessary

> In my experience they are often used for larger changes, breaking changes or changes in technical direction that have cross cutting concerns.

* A proposal should include:

** What
** Why
** How

WDYT?

README.md
secondsun: Because a proposal isn't required to be a single document.

For example if I am writing a proposal for a Android Studio plugin I would like to attach images which are mockups of what I am talking about.  The easiest way to keep them all together is to put them in a single directory.

README.md
secondsun: ON the one hand, yay structure.  On the other hand if we can't look at a proposal in a PR and answer those questions then we ask them in the PR and the proposal gets updated.

README.md
maleck13: I see no issue with either adding those images to the same directory or creating a sub dir for assets

README.md
maleck13: I think outlining what we expect from a proposal can save time and effort on our part. It doesn't have to be those heading exactly but those things should make up the contents of the proposal.
I see it as being similar to saying we expect unit tests for your pull requests, and for it to be formatted to the following standard. It sets a high level standard that should be adhered too before the PR is made

secondsun: Updated

secondsun:
A sample proposal from @maleck13 https://github.com/openshift/ansible-service-broker/blob/master/docs/proposals/last-operation-description.md

I like it especially compared to the Android SDK doc I added in pulls/2 as it actually proposes something.

README.md
secondsun: Would we want to go ahead and create some top level directories then?

At first I was thinking that proposals would be very short lived, either getting shot down or turned into trello cards and such and then deleted.  If we want them to stick around then having more curation up front like you are proposing is probably a good idea.

secondsun:
@wei-lee @maleck13 Since I will be going on PTO soon I've made you guys collaborators on my repo.  Feel free to make any edits that we don't get to today there and merge as you feel ready.

README.md
wtrocki: I think that proposals can be done as single file initially and after some time repo maintainers can just organize them into the structure they want. Less rules initially will be better. Some proposals can be retired etc. We can manage that as we go.

wtrocki:
Some good core examples for proposals that can be templates:
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/
https://github.com/nodejs/node-eps

maleck13:
@wtrocki Yep some of those are where I am pulling ideas from. I have a proposal open with K8s at the moment which is a good example of how it works and community involvement
https://github.com/kubernetes/community/pull/1166

wtrocki:
Let's merge this README. Added some minor fixes for comments. 
Anyone against that?

wtrocki (approved):

maleck13:
Personally still not a fan of the idea that we expect a new dir for every proposal.

wtrocki:
Ok. I'm going to change that.

PR #2, opened 2017-12-14
Porting Android doc from Google Docs

secondsun:
Port of a previous Android SDK document from Google Drive

### DIFF
--- a/android-sdk/README.md
+++ b/android-sdk/README.md
@@ -1,0 +1,80 @@
+MCP Android SDK Goals/Metrics
+
+# Cross Cutting Goals
+
+* Releases (mavencentral, jcenter, jitpack.io etc)
+    * Release final builds to to Maven Central
+    * Continue using [Sonatype](https://oss.sonatype.org) nexus for snapshots
+
+* CI/CD
+    * Setup auto publishing of SNAPSHOT builds
+
+* BOM style dependency management 
+    * Source a gradle plugin to support BOMs
+    * Ex [https://github.com/aerogear/aerogear-parent](https://github.com/aerogear/aerogear-parent)
+
+* Move everything to aerogear github organization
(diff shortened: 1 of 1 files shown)


wtrocki:
Proposals tend to have some specific structure and be clearly stating goal, motivation, set of steps etc. For SDK we will need to have first top level strategy around repositories etc. 
This PR can be transformed to at least 5-8 different proposals, with lots of considerations inside. 

I fully agree with all points and options, but I think that we will need to tackle the most important elements that we need for the moment and then draft proposals for the rest at later stage. 
Proposals should be achievable in sprint. Going to draft mobile strategy in separate document and work to incorporate ideas from this PR.

android-sdk/README.md
```diff
@@ -0,0 +1,80 @@
...
+    * Setup auto publishing of SNAPSHOT builds
+
+* BOM style dependency management 
+    * Source a gradle plugin to support BOMs
+    * Ex [https://github.com/aerogear/aerogear-parent](https://github.com/aerogear/aerogear-parent)
+
+* Move everything to aerogear github organization
+  * What is included in everything?
+
+* Event driven API (either promise style or listeners. NO CALLBACKS! [They are hard to use in Android]
```
josemigallas, line 19: Aren't listeners just a kind of callback?

secondsun: Yes and no.  There is a bit more formality with listeners usually than callbacks which are usually written as anonymous inner classes inline with a function reference.  

The comment is basically saying that I want function calls to lean more towards formality than toward inline callbacks.  I can whip up some pseudo code if you like to demonstrate my point more clearly.

secondsun:
closing per https://github.com/aerogear/proposals/pull/6#discussion_r159250764

PR #3, opened 2017-12-14
proposal outlining design and ux principles

maleck13:
This proposal outlines the design and ux principles that I feel work well and keep us close to those provided by the kubectl tool. Most of it is common sense.

### DIFF
--- a/cli/goals_principles.md
+++ b/cli/goals_principles.md
@@ -13,7 +13,7 @@ The key goal of the mobile cli, is to provide a simple and intuitive interface t
 
 ## Output
 
-As this is a CLI, it is important that it remains a programable grep and pipe friendly interface to allow scripting and execution from external programs. This generally means it should write its output to stdOut. Using a format like JSON allows it to be piped to tools such as [jq](https://stedolan.github.io/jq/) for further use.
+As this is a CLI, it is important that it remains a programmable grep and pipe friendly interface to allow scripting and execution from external programs. This generally means it should write its output to stdOut and its errors to stdErr. Using a format like JSON allows it to be piped to tools such as [jq](https://stedolan.github.io/jq/) for further use.
 
 However it should default to presenting its output in a human readable format such as a table structure:
 ```
@@ -32,7 +32,7 @@ However it should default to presenting its output in a human readable format su
 ```
 mobile get serviceinstances fh-sync-server --output=json
 ```
-In this example we have  ```servicesinstances``` and ```fh-sync-server``` that are required but the ```--output``` flag is optional to change the default behaviour for how the output is formatted.
+In this example we have  ```servicesinstances``` and ```fh-sync-server``` that are required but the ```--output``` flag is optional to change the default behaviour for how the output is formatted. As much as possible we want to align with the kubectl practices here to keep it a familiar interface.
 
 ## Errors
 
--- a/cli/goals_principles.md
+++ b/cli/goals_principles.md
@@ -1,0 +1,74 @@
+# Abstract
+
+This document will outline the goals of creating a mobile CLI for K8s / OpenShift. It will also outline the technical considerations and design principles that should be used to guide the development of not only the mobile CLI, but any other mobile service CLI.
+
+
+
+# Goals
+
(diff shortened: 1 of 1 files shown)


cli/goals_principles.md
```diff
@@ -0,0 +1,74 @@
...
++-----------------------+-----------------------------+
+|         NAME          |             ID              |
++-----------------------+-----------------------------+
+| dh-fh-sync-server-apb | dh-fh-sync-server-apb-59dkt |
++-----------------------+-----------------------------+
+```
+
+## Arguments and Flags
+	
+- An argument represents a required parameter needed in order to complete the requested action.
```
philbrookes, line 29: Going by the example, is it worth mentioning that arguments have a required order too?

cli/goals_principles.md
```diff
@@ -0,0 +1,74 @@
...
+- A flag represent an optional value needed for additional functionality or to change default behaviour. Flags can be global, as in spanning all commands or specific to a single command. Were appropriate they should have a short and long format.
+
+```
+mobile get serviceinstances fh-sync-server --output=json
+```
+In this example we have  ```servicesinstances``` and ```fh-sync-server``` that are required but the ```--output``` flag is optional to change the default behaviour for how the output is formatted.
+
+## Errors
+
+In the case of an error, there should be a clear message printed as the first line of output followed by the usage for that command, the process should then exit with a non zero exit code so then when being called from an external program it is clear when something has failed.
```
philbrookes: I think this output should come out on stderr also, to stay compatible with typical CLI tools.

philbrookes: Do you think it might be better to output the human-readable error under the usage instead, to reduce the chance it get's pushed out of the top of the console window?

cli/goals_principles.md
maleck13: No I think that is covered under the user instruction section

cli/goals_principles.md
maleck13: Yeah agree missed that

maleck13: personally no. but I think that perhaps there is an argument for not showing the usage and only showing the error.

cli/goals_principles.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Arguments and Flags
+	
+- An argument represents a required parameter needed in order to complete the requested action. 
+- A flag represent an optional value needed for additional functionality or to change default behaviour. Flags can be global, as in spanning all commands or specific to a single command. Were appropriate they should have a short and long format.
+
+```
+mobile get serviceinstances fh-sync-server --output=json
+```
+In this example we have  ```servicesinstances``` and ```fh-sync-server``` that are required but the ```--output``` flag is optional to change the default behaviour for how the output is formatted.
```
wtrocki: I think it's worth to mention that this aligns with oc cli (and we may just align with format to oc so people who use oc will feel natural about cli)

cli/goals_principles.md
```diff
@@ -0,0 +1,74 @@
...
+
+Flags:
+  -h, --help      help for serviceinstance
+      --no-wait   --no-wait will cause the command to exit immediately instead of waiting for the service to be provisioned
+
+Global Flags:
+      --namespace string   --namespace=myproject
+  -o, --output string      -o=json -o=template (default "table")
+
+error: exit status 1
```
wtrocki, line 55: This looks like standard format from [cobra](https://github.com/spf13/cobra) Did you made any non standard modifications?

maleck13: Nope cobra is the same tool used for kubectl and oc so seemed like a good standard to go by

maleck13:
@philbrookes @wtrocki updated based on your feedback. If you are happy could you approve the PR

wtrocki:
I have played a little bit with client. 
Really like that we document objects and provide operations on them - same as openshift.  https://github.com/aerogear/mobile-cli#core-objects-or-resources

wtrocki (approved):
Read document again. Looks great!

philbrookes (approved):

PR #4, opened 2017-12-15
proposal for creating an additinal configmap when provisioning a service

maleck13:
Outlines a change to what the APBs should create when provisioning and deprovision services

### DIFF
--- a/apbs/create-secret-and-configmap-during-provision.md
+++ b/apbs/create-secret-and-configmap-during-provision.md
@@ -4,7 +4,7 @@ This document outlines a proposal to create an additional configmap for services
 
 ## Terms
 
-- Mobile Client (a mobile device in the public domain running a developers application)
+- Mobile Client (a mobile device running a developers application)
 - APB (ansible playbook bundle)
 
 ## Problem Description
@@ -16,7 +16,7 @@ Our current approach is to create a secret during the provision of a new mobile 
 When provisioning a new service instance, we should create two configuration resources:
 
 1) A secret containing private information such as usernames and passwords or anything that we do not want to be exposed to the mobile client.
-2) A configmap containing all of the required information needed for a mobile client to access the service.
+2) A configmap containing all of the required information needed for a mobile client to use the service.
 
 Note it may be neccessary to duplicate some information across each of these resources (ie the URI of the service for example).
 
--- a/apbs/create-secret-and-configmap-during-provision.md
+++ b/apbs/create-secret-and-configmap-during-provision.md
@@ -1,0 +1,44 @@
+# Abstract
+
+This document outlines a proposal to create an additional configmap for services accessible via a mobile client, this configmap would compliment the secret we currently create during the provision of each service via an APB 
+
+## Terms
+
+- Mobile Client (a mobile device in the public domain running a developers application)
+- APB (ansible playbook bundle)
+
+## Problem Description
+
+Our current approach is to create a secret during the provision of a new mobile service. This secret contains all of the information required to allow the UI to show the service and how to access the service, it also includes any information needed to create the mobile client configuration. This means we expose too much information and also need to filter out any sensitive information (usernames password etc) when constructing our mobile configuration to avoid it being exposed to the client and ending up in the public domain. As each service is different, it often requires a custom piece of code to filter out the fields that we don't want or need exposed. 
+
+## Proposed Solution
+
+When provisioning a new service instance, we should create two configuration resources:
+
+1) A secret containing private information such as usernames and passwords or anything that we do not want to be exposed to the mobile client.
+2) A configmap containing all of the required information needed for a mobile client to access the service.
+
+Note it may be neccessary to duplicate some information across each of these resources (ie the URI of the service for example).
+
+Each of these resources should have at least the following information and labels:
+
+- Labels: 
+    - **mobile:enabled**. Will give a mechanism to filter these resources from other resources of the same type from the UI and CLI
+    - **serviceName:< serviceName >**. Depending on whether we want to allow more than one service of a particular kind to be provisioned to a namespace, this will allow us to identify which service the information is for without needing to name the resource itself after the service.
+
+- Data
+    - **URI** both the configmap and secret will require this. 
+
+
+### CLI and UI changes
+
+Currently the CLI and UI look for secrets to create the mobile client config, this will need to change to look for configmaps with the label ```mobile:enabled```.
+There should be no changes to representing the services as the secret will continue to exist.
+
+
+### APB Changes
+
+Each APB for a service that needs to be exposed to the mobile client, will need to be changed to create the specified configmap. The APB will also need to be changed to remove this configmap when the service is removed.
+
+
+


philbrookes (approved):
:+1:

apbs/create-secret-and-configmap-during-provision.md
```diff
@@ -0,0 +1,44 @@
+# Abstract
+
+This document outlines a proposal to create an additional configmap for services accessible via a mobile client, this configmap would compliment the secret we currently create during the provision of each service via an APB 
+
+## Terms
+
+- Mobile Client (a mobile device in the public domain running a developers application)
```
david-martin: 'public domain' has a particular meaning around intellectual property.
Can probably just leave it out
`a mobile device running a developers application`

apbs/create-secret-and-configmap-during-provision.md
```diff
@@ -0,0 +1,44 @@
...
+- APB (ansible playbook bundle)
+
+## Problem Description
+
+Our current approach is to create a secret during the provision of a new mobile service. This secret contains all of the information required to allow the UI to show the service and how to access the service, it also includes any information needed to create the mobile client configuration. This means we expose too much information and also need to filter out any sensitive information (usernames password etc) when constructing our mobile configuration to avoid it being exposed to the client and ending up in the public domain. As each service is different, it often requires a custom piece of code to filter out the fields that we don't want or need exposed.
```
david-martin, line 12: The 'Service Instance' created by the catalog is probably a better thing to determine if the service should be shown in the UI.
This is already the case on the Overview screen in the 'Provisioned Services' section.
Mobile services are likely to show up in that section too, but have some additional UI when you expand it

apbs/create-secret-and-configmap-during-provision.md
```diff
@@ -0,0 +1,44 @@
...
+
+1) A secret containing private information such as usernames and passwords or anything that we do not want to be exposed to the mobile client.
+2) A configmap containing all of the required information needed for a mobile client to access the service.
+
+Note it may be neccessary to duplicate some information across each of these resources (ie the URI of the service for example).
+
+Each of these resources should have at least the following information and labels:
+
+- Labels: 
+    - **mobile:enabled**. Will give a mechanism to filter these resources from other resources of the same type from the UI and CLI
```
david-martin, line 26: `mobile:true` might be better as it could be parsed into a boolean easier if needed. Also, is it something that can be 'disabled'? A boolean says it is or isn't something rather than something that's enabled or disabled.

apbs/create-secret-and-configmap-during-provision.md
```diff
@@ -0,0 +1,44 @@
...
+
+### CLI and UI changes
+
+Currently the CLI and UI look for secrets to create the mobile client config, this will need to change to look for configmaps with the label ```mobile:enabled```.
+There should be no changes to representing the services as the secret will continue to exist.
+
+
+### APB Changes
+
+Each APB for a service that needs to be exposed to the mobile client, will need to be changed to create the specified configmap. The APB will also need to be changed to remove this configmap when the service is removed.
```
david-martin, line 41: Probably a separate proposal, but should APB deprovision leverage labels that we've added so they remove all relevant resources matching that label?

apbs/create-secret-and-configmap-during-provision.md
```diff
@@ -0,0 +1,44 @@
...
+Our current approach is to create a secret during the provision of a new mobile service. This secret contains all of the information required to allow the UI to show the service and how to access the service, it also includes any information needed to create the mobile client configuration. This means we expose too much information and also need to filter out any sensitive information (usernames password etc) when constructing our mobile configuration to avoid it being exposed to the client and ending up in the public domain. As each service is different, it often requires a custom piece of code to filter out the fields that we don't want or need exposed. 
+
+## Proposed Solution
+
+When provisioning a new service instance, we should create two configuration resources:
+
+1) A secret containing private information such as usernames and passwords or anything that we do not want to be exposed to the mobile client.
+2) A configmap containing all of the required information needed for a mobile client to access the service.
+
+Note it may be neccessary to duplicate some information across each of these resources (ie the URI of the service for example).
```
david-martin: Is it likely the URI will need to be updated, and hence updated in 2 places?

apbs/create-secret-and-configmap-during-provision.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Problem Description
+
+Our current approach is to create a secret during the provision of a new mobile service. This secret contains all of the information required to allow the UI to show the service and how to access the service, it also includes any information needed to create the mobile client configuration. This means we expose too much information and also need to filter out any sensitive information (usernames password etc) when constructing our mobile configuration to avoid it being exposed to the client and ending up in the public domain. As each service is different, it often requires a custom piece of code to filter out the fields that we don't want or need exposed. 
+
+## Proposed Solution
+
+When provisioning a new service instance, we should create two configuration resources:
+
+1) A secret containing private information such as usernames and passwords or anything that we do not want to be exposed to the mobile client.
```
david-martin, line 18: Giving an example here might help.

e.g. the administrator username & password

apbs/create-secret-and-configmap-during-provision.md
```diff
@@ -0,0 +1,44 @@
...
+2) A configmap containing all of the required information needed for a mobile client to access the service.
```
david-martin: would 'use' sound better than 'access'

david-martin (changes requested):

apbs/create-secret-and-configmap-during-provision.md
maleck13: Sure however the service instance wont have the credentials and additional info right? So the UI will still need to read the secret?

apbs/create-secret-and-configmap-during-provision.md
maleck13: I do mention usernames and passwords? Not sure we need more than that

apbs/create-secret-and-configmap-during-provision.md
david-martin: Why does the UI need to read the secret? Are we showing the credentials?
Also, what additional info is there that the UI should show?

We can certainly look up the secret linked/created for the service instance if/when needed

apbs/create-secret-and-configmap-during-provision.md
maleck13: I would think that would be a rare event

apbs/create-secret-and-configmap-during-provision.md
maleck13: the label is just an identifier a way of filtering. Also labels can only be strings so cant really use true false unless they are parsed in the client and then there is no difference to using enabled

apbs/create-secret-and-configmap-during-provision.md
maleck13: I think the key thing is that the deprovision cleans up everything, how that is actually done by the APB is probably more of a implementation detail

apbs/create-secret-and-configmap-during-provision.md
david-martin: I must have misread that line and missed the mention of username & password. Can ignore

maleck13:
@david-martin updated with the minor changes.

apbs/create-secret-and-configmap-during-provision.md
david-martin: :+1:

apbs/create-secret-and-configmap-during-provision.md
david-martin: OK, only strings makes sense then. JSON.parse or equivalent on a client would parse "true" to a boolean

maleck13: Really? That seems bad. In Javascript maybe it does that. Pretty sure that wouldn't work in Go though

apbs/create-secret-and-configmap-during-provision.md
maleck13: I may be out of the loop on the UX but originally we had been showing a screen where the url and login credentials were shown to allow a developer to jump into the service?

david-martin: A good thing to discuss with UX as it hasn't been discussed yet. Just the idea of a link i.e. 'Dashbaord URL' showing up was discussed
cc @cshinn.

Either way, reading the secret would/should only be needed for that (if the user has perms to read the secret).
Reading the Service Instance to populate the 'Provisioned Services' section (as it does now) make sense to me

maleck13: so maybe a separate proposal to replace using the secret with using the ServiceInstance? Either way it is not a  major part of this proposal

apbs/create-secret-and-configmap-during-provision.md
```diff
@@ -0,0 +1,44 @@
...
+## Problem Description
+
+Our current approach is to create a secret during the provision of a new mobile service. This secret contains all of the information required to allow the UI to show the service and how to access the service, it also includes any information needed to create the mobile client configuration. This means we expose too much information and also need to filter out any sensitive information (usernames password etc) when constructing our mobile configuration to avoid it being exposed to the client and ending up in the public domain. As each service is different, it often requires a custom piece of code to filter out the fields that we don't want or need exposed. 
+
+## Proposed Solution
+
+When provisioning a new service instance, we should create two configuration resources:
+
+1) A secret containing private information such as usernames and passwords or anything that we do not want to be exposed to the mobile client.
+2) A configmap containing all of the required information needed for a mobile client to use the service.
```
matzew, line 19: for push, the auth for device registration (done via HTTP_BASIC) would be needed in that configmap - since the client has to know that variantID:secret (aka user/password)

maleck13: yeah I think if it is needed by the client then it should go in the configmap

apbs/create-secret-and-configmap-during-provision.md
david-martin: agreed

apbs/create-secret-and-configmap-during-provision.md
david-martin: Yeah, e.g
```
typeof JSON.parse('true')
'boolean'
```
anyways, :+1: on this

apbs/create-secret-and-configmap-during-provision.md
david-martin: I'm good with a separate proposal for that

david-martin (approved):

apbs/create-secret-and-configmap-during-provision.md
```diff
@@ -0,0 +1,44 @@
...
+Note it may be neccessary to duplicate some information across each of these resources (ie the URI of the service for example).
+
+Each of these resources should have at least the following information and labels:
+
+- Labels: 
+    - **mobile:enabled**. Will give a mechanism to filter these resources from other resources of the same type from the UI and CLI
+    - **serviceName:< serviceName >**. Depending on whether we want to allow more than one service of a particular kind to be provisioned to a namespace, this will allow us to identify which service the information is for without needing to name the resource itself after the service.
+
+- Data
+    - **URI** both the configmap and secret will require this.
```
wei-lee, line 30: for services that will have other config options, I think it is better to group them together and give them a key called `config`, rather than having them at the root of the configmap object.

e.g.

```
{
  mobile: 'enabled',
  serviceName: 'push',
  URI: 'https://...',
  config: {
    key:"...",
    secret: "...."
  }
}
```

maleck13: The key values should all be configuration in the configmap, and the labels then (which wont show up in the config objects) give the meta data.

wei-lee: also probably each each config should have the same format, e.g. each extra config should have `name` and `value` fields. This is to make it easier to lookup and manage the config options.

PR #5, opened 2017-12-19
[WIP] Proposal for metrics with Prometheus & Grafana

david-martin:


david-martin:
@pb82 This might be useful to be aware of

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
```diff
@@ -0,0 +1,136 @@
...
+
+If a Service is exposing a HTTPS metrics endpoint at `/rest/metrics`, the annoation would be:
+
+```yaml
+org.aerogear.metrics/ssl_endpoint: /rest/metrics
+```
+
+### Default Visualisations
+
+Grafana will be configured 'out of the box' with 'Dashboards' for specific Mobile Services. Which services will be determined on a case by case basis. These Dashboards will be tailored to the needs of a developer using that service. For example, the Data Sync Dashboard will show queue metrics.
```
philbrookes: do you think there's any promise in the idea of configuring a dashboard when the prometheus and <service> are bound?
It'd mean we wouldn't have a bunch of blank dashboards, but would also mean that the user would have to bind the services if they want to see the dashboard for that service.

david-martin: What are your thoughts on this as an approach instead:

* Grafana comes with no dashboards by default
* We include/write a new Grafana plugin that looks for metadata about dashboards that Services in your namespace have defined
* This plugin updates the list of available dashboards in Grafana to include those learned from Services metadata 

For example, the keycloak service could include a configmap or an annotation with the raw dashboard json content. The plugin is regularly scanning for any new resources in the namespace that have this content. It then picks up on the raw dashboard json content for keycloak & adds it to the list of available dashboards.

I'm currently assessing the feasibility of this approach.
Alternatively, a sidecar approach could be used where the sidecar monitors namespace resources, and updates the dashboard configmap accordingly.

FYI, updating the dashboard configmap currently will automatically update mounted files (including adding any new files) in the grafana container. Grafana also automatically updates the list based on available dashboard files in the configured location.

david-martin: A 'pro' for above approach is the developer will always get the dashboard that is relevant to their version of the service. Also, if they upgrade their service (e.g. ups), the metadata would also get updated, and their dashboard also as a result.

They are free to 'copy' the dashboard and make changes to it, but the read-only ones that show up will always be the ones that shipped (and were tested with) with the service version they are running.

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
```diff
@@ -0,0 +1,136 @@
...
+The Metrics Service will consist of:
+
+* Prometheus - for metrics gathering
+* Grafana - for metrics visualisation
+
+Promtehues & Grafana will each have a Persistent Volume for storing gathered metrics and visualisation settings respectively.
+
+Grafana will be configured to have Proemtheus available as a 'Data Source' by default. Data Sources will be stored in a ConfigMap that gets mounted into the Grafana container. e.g.
+
+https://github.com/aerogearcatalog/prometheus-apb/blob/ee86cfc67caaeaabcf7e7bc214c3a6dc6940c333/roles/provision-prometheus-apb/tasks/main.yml#L25-L26
```
matzew: should we name it different than `prometheus-apb` ?

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
```diff
@@ -0,0 +1,136 @@
...
+* Service specific logic required to be written and maintained in the MCP UI
+* Memory footprint of the `mcp-standalone` server was indeterminate
+
+As the number of Mobile Services and supported other Services increases, the size and maintenance overhead of the server & UI would be very large.
+
+Also, post proof of concept, it was decided the `mcp-standalone` server would not be continued as a concept, or any other 'required' server component specific to Mobile.
+
+## Proposed Solution
+
+### Metrics Service
```
matzew, line 52: Prometheus is tech preview in Openshift (not sure about Grafana)

would it make sense to leverage what the platform itself offers ?

david-martin: Absolutely, if possible.
Is there any info we have yet about what hooks we could get into a cluster metrics with prometheus, or if a grafana could use it as a data source?

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: `metrics-apb`?

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: This may prove useful for a sidecar approach https://github.com/PierreVincent/k8s-grafana-watcher. Would be good to evaluate if we want to go that route.

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
```diff
@@ -0,0 +1,136 @@
...
+  url: http://localhost:9090
+  is_default: true
+  version: 1
+  editable: true
+  json_data: {"timeInterval": "5s"}
+```
+
+### Metrics Discovery
+
+Prometheus will be configured to 'auto-discover' any Services in the same Kubernetes namespace that expose a Prometheus style metrics endpoint. This will be done via an annoation on the `Service` resource.
```
wei-lee, line 81: I can't remember if this has been shown in the demo, but can developers drill into the metrics data for each pod, or they can only get aggregated metrics from all the pods that are behind a service?

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
wei-lee: +1 on the sidecar approach. I think it will be a much better user experience if we can do it

wei-lee: I also like the idea of having each service provides its ow raw Grafana dashboard content. No need to worry about version compatibility between the dashboards and the services.

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
```diff
@@ -0,0 +1,136 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: proxy
+  annotations:
+    serviceaccounts.openshift.io/oauth-redirectreference.grafana: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"grafana"}}'
+    serviceaccounts.openshift.io/oauth-redirectreference.prometheus: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"prometheus"}}'
+```
+
+Both OAuth Proxies, Grafana & Prometheus will all run in the same Pod, but in separate containers. This is necessary so Grafana can talk directly to Promethues via localhost, bypassing the OAuth Proxy when accessing it as a Data Source. Also, the OAuth Proxies can proxy directly on localhost to Grafana & Prometheus without having to expose a Service, making the unprotected services potentially accessible outside the namespace.
```
wei-lee: > Both OAuth Proxies, Grafana & Prometheus will all run in the same Pod, but in separate containers. This is necessary so Grafana can talk directly to Promethues via localhost, bypassing the OAuth Proxy when accessing it as a Data Source. 

Will this make it hard to scale the metrics service?

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
philbrookes: yeah, I love this solution, :+1:

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
philbrookes: `metrics-apb` sounds good to me.

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
philbrookes: as it stands, the user can only get information based on what is exposed on the metrics end-point. I do not believe there would be a way for the metrics end-point to reveal data about specific pods. This might be an area where cadvisor is more useful?

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: "To put some numbers behind this: a single big Prometheus server can easily store millions of time series, with a record of 800,000 incoming samples per second (as measured with real production metrics data at SoundCloud). Given a 10-seconds scrape interval and 700 time series per host, this allows you to monitor over 10,000 machines from a single Prometheus server"
https://prometheus.io/blog/2016/07/23/pull-does-not-scale-or-does-it/
I say lets cross that bridge when we get to it as it will likely need a completely different approach

wei-lee: @david-martin  I am not talking about how many machines a single Prometheus server can support, I am talking about having Grafana & Prometheus running in the same pod. In this setup, we can not scaling Grafana without scaling Prometheus - but as you said, a single Prometheus can support many machines, so scaling Prometheus seems unnecessary. Feels like it's better to keep them in separate pods, and figure out how to make Grafana authenticate with Prometheus?

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
```diff
@@ -0,0 +1,136 @@
+# Abstract
+
+This document outlines a proposal for gathering metrics from Mobile Services running in a Kubernetes environment and visualizing them.
+
+## Goals
+
+### Primary Goals
+
+* Provide a mechanism for Mobile Developers to gather, view & query metrics from supported Mobile Services.
```
wtrocki: I might be getting this wrong but saying "mobile developers" is misleading and my suggest mobile metrics. I think it's best to rename `mobile` to "DevOps" or "Server side" developers. Considering non goals bellow there is no specific mobile focus and it can be simply added as "openshift metrics service"

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
```diff
@@ -0,0 +1,136 @@
...
+### Metrics Service
+
+No central metrics gathering or visualisations will be available for Mobile Services 'out of the box'. Instead, if a developer wants metrics, they can provision a Metrics Service from the Service Catalog.
+
+The Metrics Service will consist of:
+
+* Prometheus - for metrics gathering
+* Grafana - for metrics visualisation
+
+Promtehues & Grafana will each have a Persistent Volume for storing gathered metrics and visualisation settings respectively.
```
wtrocki: Typo in `Promtehues`

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: @wei-lee Good point. Scaling Grafana may be something the user wants to do.
From doing some reading, it should be possible to have a shared filesystem/persistent volume for grafana storoage of plugins & dashboards
https://medium.com/@fcgravalos/scaling-out-grafana-with-kubernetes-and-aws-62745257df10

1) Solving the auth problem between Grafana & Prometheus could be tricky. To keep the oauth-proxy in front of prometheus, grafana would need to have an oauth-proxy cookie header (which is got by logging in via oauth). I'm not sure how that could be done.

2) We could remove the oauth-proxy in front of prometheus, but we'd have to remove the exposed route too. Then Grafana could access it on the service name without auth i.e. 'prometheus'. You wouldn't be able to access the prometheus dashboard.

3) Another option could be to expose another service in front of prometheus that bypasses the oauth-proxy. This service would only be internally accessible/routable. The exposed external route to prometheus would still be routed through the oauth proxy.
However, I'm not sure if having the extra service would protect it from being accessed from another namespace via the full service host i.e. prometheus.namespace.svc.cluster.local

david-martin: Option 3 could be a runner if projects have isolated networks
https://docs.openshift.com/container-platform/3.7/admin_guide/managing_networking.html#joining-project-networks
@philbrookes any thoughts on this? I think we discussed it before but I may have ruled it out due to security concerns about accessibility from other projects. I believe that shouldn't be the case though

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: @wei-lee @philbrookes I'll reword this section of the proposal based on above discussion

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: I'll make a note in the proposal to use that name

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
pb82: @david-martin @matzew there is https://github.com/openshift/origin-metrics but it looks like it's designed to only gather pod and container metrics provided by a `/stats` endpoint, not custom metrics like queue data in sync etc.

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: Good point. I can rephrase.
How does this sound for the primary

"Provide a mechanism for developers to gather, view & query metrics from supported Mobile Services"

and the secondary

"Allow Developers to leverage the same mechanism for any other/non-mobile/custom services"

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: @pb82 There is also https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_prometheus#openshift-prometheus (tech preview in 3.7). However, I don't know if it will be possible to specify custom metrics endpoints per namespace to include in a cluster level prometheus. I'll see what I can find out and ask some questions

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
```diff
@@ -0,0 +1,136 @@
...
+
+```yaml
+org.aerogear.metrics/plain_endpoint: /metrics
+```
+
+If a Service is exposing a HTTPS metrics endpoint at `/rest/metrics`, the annoation would be:
+
+```yaml
+org.aerogear.metrics/ssl_endpoint: /rest/metrics
+```
```
wei-lee, line 93: @david-martin @philbrookes  Does Prometheus support user authentication when pull the metrics data from those services? If not, does that mean the metrics endpoints will be public available?

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: For Option 1, using the `--skip-auth-regex` flag, as in the PR below could allow us to bypass the oauth proxy for certain urls
https://github.com/openshift/origin/pull/17114/files#diff-1630f01082dce4d5ef4ba69488c022a8R125

The datasource definition in the same PR uses a serviceaccount token too
https://github.com/openshift/origin/pull/17114/files#diff-cd98aab0c9f8f84ad1e48d348b459d0fR53

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: The `scrape_config` can specify params, basic auth, Authorization token (bearer) and some other things.
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#%3Cscrape_config%3E
As we do a service integration (e.g. metrics for keycloak) we'll have to decide which auth type makes sense, and how to auto-configure it.
@philbrookes Would you agree? Is it possible to specify the scrape_config with our current approach?

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
wtrocki: @david-martin Lists metrics are available out of the box:
https://github.com/openshift/origin/tree/master/examples/prometheus

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
philbrookes: At the moment the scrape config is contained in a template file in the prometheus-apb. We can change that, but it would change for all services that match the service-discovery flag.
We could add some more annotations to the services to be discovered, that would allow them to specify a username and password which Prometheus should use to read the metrics end-point, in the same way we specify what the metrics end-point is - and have it default to unsecured if they are not present?

wei-lee: Not sure I like the idea of using annotations to store username/password, mainly because of security concerns. Can we create a secret and mount it to both the service & prometheus?

david-martin: @wei-lee From chatting with @philbrookes and doing some reading, I don't think that's possible out of the box. It may require changes to Prometheus, or a sidecar approach that dynamically generates the Prometheus config instead of leveraging Prometheus relabeling.

@philbrookes is investigating upstream on this.
A first pass could be to pass an apikey param, as param relabeling is supported via `__param_<name>`. The metrics endpoint in the service would then check the apikey matches.

More info in the docs here https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: Updating to go with Option 3.

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: @wei-lee If the service exposing the metrics endpoint is gathering metrics for cpu & memory (based on what it can observe inside the container), then they can show that.
However, I would see Pod metrics like that being provided by the cluster metrics (heapster atm, a different prometheus instance probably in future) as that would have the necessary perms to read cgroups/cadvisor stuff, and be more accurate

david-martin:
I've pushed some minor updates (typos & separate Pod changes for scaling).
Holding off on updates regarding Grafana Dashboards solution & metrics endpoint security, pending  internal discussion/knowledge sharing with another group.

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
wtrocki: @david-martin It maybe outside of the scope of this proposal, but are we going to support metrics collection outside cluster (for example from mobile)?
Is it fair to say that metrics will be used only for cluster wide metrics?

This is follow up to comment I have made here: 
https://trello.com/c/lt4jc3h4/10-as-a-developer-i-want-to-be-able-to-visualise-runtime-metrics-about-what-versions-of-the-mobile-service-sdks-are-in-use-so-that

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
aliok: I've got something similar:

If I would like to track a metric like "number of requests to /foo endpoint", is there a way to handle this without an external cache like Redis?

I think in this case, normally one would increase some counter in a Redis instance which is used by all pods in a service. Then the metrics endpoint would just read the counter from Redis.

However, I am just trying to investigate a way like:
* each pod tracks that metric in memory
* prometheus visits each pod to get the metric value
* I show an aggregation of the metric value in Grafana. Or, I show requests per pod, although this is not very useful in this case.
* Keypoint: **no Redis**

david-martin: @aliok it is possible to look for pod annotations for metrics endpoint discovery. However, to date we've only used service annotations
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config

What your suggesting sounds feasible thouugh, but I haven't tried it. I'm not sure if prometheus allows metrics endpoints across multiple pods to return values for the same metrics keys. Maybe that doesn't matter if you are just summing up the total on your graph, and your query knows what the keys are

@philbrookes thoughts?

aliok: @david-martin ok, sounds good. I think using pod annotations would cover that scenario. I was just trying to understand the limits of the thing. no further investigation is needed.

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: https://issues.jboss.org/browse/AEROGEAR-1819

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: After doing some research, a discussion thread, and a call, it might be possible in future to use the the cluster prometheus for monitoring application metrics, but there would be strict limits on it, and probably only available via some auto discovery (not configurable by users).
This would be for scale and security reasons.

Having a separate Prometheus for application/user metrics is the route we, and other RH projects, will be taking

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: I'm going to leave authentication out of the proposal at this time, but pass it on as a use-case to the team maintaining the Prometheus image.

metrics/prometheus-metrics-endpoints-and-auto-discovery.md
david-martin: Reworded this section to follow a simpler approach for now. Investigation issue being created to look into auto discovery

PR #7, opened 2018-01-04
Using asciibinder for Aerogear.org 

pwright:


aerogear-docs.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+This demo uses github and circleci, fork and build as required. 
+
+Consider the following 'feature' doc repos:
+
+* https://github.com/finp/sync-docs
+* https://github.com/finp/push-docs
+* https://github.com/finp/digger-docs
+
+Each release branch be rendered using asciibinder, for example, see link:https://5-114535426-gh.circle-artifacts.com/0/home/circleci/docs/_preview/digger/latest/installation/digger-install-intro.html[Digger docs].
```
david-martin: Minor grammar thing here.
'Each release branch be' => 'Each release branch will/should be'

david-martin:
I like this idea.
If the retrieval of docs from x many other repos can be kept automated, that would be great.

Maybe outside the scope of this proposal, but should we be thinking about how search will work on the site?

aerogear-docs.adoc
pwright: thanks, it was a bit early for 'write like a pirate' day

aerogear-docs.adoc
```diff
@@ -0,0 +1,104 @@
+= Aerogear.org Website Renewal
```
wtrocki: `Aerogear.org modular documentation` seems to be better tittle ;)

aerogear-docs.adoc
```diff
@@ -0,0 +1,104 @@
...
+
+== Summary
+Everyone should be able to contribute to aerogear.org docs with ease and effectiveness, regardless of the scope of their proposed change.
```
wtrocki: @finp I love this proposal! Really good job. 
We added some simple template to follow (mostly headers) that will help with verification and approving this proposal. See: https://github.com/aerogear/proposals/blob/master/template.md

wtrocki:
It think is worth to mention that proposal it's only to cover documentation for services. 
Other website elements are not covered.

david-martin:
@finp We have something tangible now with https://github.com/aerogear/service-docs.
This is a single repo for anything related to the mobile services that end users (Mobile/Mobile Backend developers) would find useful.

I'd imagine there'd be other docs sources for things like UI guides, CLI installation & guides, IDE plugin installation & guides in the future too.

Is this OK with you?
IS there anything needing updating in the proposal based on this approach?

pwright:
@david-martin  I see the content in services-docs as being orphaned for now. We'll need a lot more context and structure before publication, but I don't see anything to block progress. I wonder where docs related to mobile-core will live? another repo? would a mobile-docs repo make sense for all generic material?

david-martin:
> would a mobile-docs repo make sense for all generic material?

Do you mean move existing content from service-docs to a new metrics-docs repo?
I don't mind if that would make things easier for you.

Approving this PR as is though

david-martin (approved):

david-martin:
Good to merge if we get another +1 or approval

wtrocki (approved):
Love idea. Demo looks great. 
Proposal is properly drafted and covers main issues.

wei-lee:
+1

wei-lee (approved):

PR #8, opened 2018-01-05
Proposal: Single Sign for Developers across all Services

david-martin:
# Abstract

This document outlines a proposal for Single Sign On across Mobile Services & Mobile Enabled Services provisioned on OpenShift. This proposal is *not* intended as a feature for end users of Mobile Apps. It *is* intended as a development time feature for Mobile developers.

### DIFF
--- a/auth/developer-single-sign-on-across-mobile-services.md
+++ b/auth/developer-single-sign-on-across-mobile-services.md
@@ -1,6 +1,6 @@
 # Abstract
 
-This document outlines a proposal for Single Sign On across Mobile Services & Mobile Enabled Services provisioned on OpenShift. This proposal is *not* intended as a feature for end users of Mobile Apps. It *is* intended as a development time feature for Mobile developers.
+This document outlines a proposal for single sign-on across Mobile Services & Mobile Enabled Services provisioned on OpenShift. This proposal is *not* intended as a feature for end users of Mobile Apps. It *is* intended as a development time feature for Mobile developers.
 
 ## Terms
 
@@ -26,7 +26,7 @@ Additionaly, a service that has no auth mechanism, e.g. Prometheus, will need to
 
 ## Proposed Solution
 
-### OpenShift as the Single Sign On Provider
+### OpenShift as the Single Sign-On Provider
 
 All Services will be running on OpenShift, or 'ordered' from the OpenShift Catalog. OpenShift is the only constant between the Developer and the Services. The Developer will need to be signed in to OpenShift. Leveraging this fact and having OpenShift be the single auth provider for all Services will address the main points in the problem description.
 
--- a/auth/developer-single-sign-on-across-mobile-services.md
+++ b/auth/developer-single-sign-on-across-mobile-services.md
@@ -1,0 +1,59 @@
+# Abstract
+
+This document outlines a proposal for Single Sign On across Mobile Services & Mobile Enabled Services provisioned on OpenShift. This proposal is *not* intended as a feature for end users of Mobile Apps. It *is* intended as a development time feature for Mobile developers.
+
+## Terms
+
+* Mobile Services - Services that provide mobile specific functionality for Mobile Apps at development time or runtime e.g. Push Notifications, Data Sync, CI/CD
+* Mobile Enabled Services - Services that provide functionality for Mobile solutions at development time or runtime, but functionality is not specific to Mobile e.g. Keycloak, Metrics.
+
+## Problem Description
+
+As a developer builds up their Project by provisioning different services, they will likely need to setup or configure ascpects of those services. For example, if a developer provisions the Push Notifications service, they will need to configure Push credentials for Mobile Apps. Some services may also provide additional info or dasboards. Accessing these dashboards & performing setup or configuration will likely require signing into those services via some auth mechanism.
+
+Each service will have 1 or more of the following:
+
+* it's own auth mechanism
+* ability to configure an external auth mechanism
+* no auth mechanism
+
+This could cause an inconsistent and poor user experience for the following reasons:
+
+* different auth flow for each service
+* different auth credentials for each service, managed independently
+
+Additionaly, a service that has no auth mechanism, e.g. Prometheus, will need to be secured in some manner if exposed externally.
+
+## Proposed Solution
+
+### OpenShift as the Single Sign On Provider
+
+All Services will be running on OpenShift, or 'ordered' from the OpenShift Catalog. OpenShift is the only constant between the Developer and the Services. The Developer will need to be signed in to OpenShift. Leveraging this fact and having OpenShift be the single auth provider for all Services will address the main points in the problem description.
+
+The flow for signing in to each Service will be the same i.e. an OAuth flow where the user signs in & authorises the Service with OpenShift. No additional credentials will be required.
+
+For cases where no auth mechanism is provided with a service, a reverse proxy approach that uses OpenShift OAuth, as shown in Option 3 below, can be used.
+
+### Provider Integration Options
+
+Choosing how to integrate a Service with OpenShift OAuth will depend on what the Service is capable of already, and how extensible it is. The below options show practical examples of how this integration was done. There will likely be tweaks and other options required depending on the service. These options serve as references for how it could be done.
+
+#### Option 1: Configure OAuth in the Service
+
+If a Service has the ability to an OAuth provider, use it. For example, in Keycloak it is possible to use 'Identity Brokering' to defer to OpenShift Auth for who can administer that Keycloak instance. See this [Keycloak Identity Brokering Blog Post](https://developers.redhat.com/blog/2017/12/06/keycloak-identity-brokering-openshift/) for details.
+
+However, with Keycloak this method is limited to only automatically setting up the initial permissions of *all* users who can successfully authenticate. There is no permissions mapping between OpenShift roles & Keycloak administration roles. Additionally, *any* OpenShift user will be able to sign in to Keycloak. Manual configuration is required to only allow certain users access and limit their roles.
+
+#### Option 2: Use a plugin for OAuth
+
+If a Service allows plugins, it may be possible to use a plugin for auth. For example, the [Jenkins OpenShift Login Plugin](https://github.com/openshift/jenkins-openshift-login-plugin#openshift-login) allows OpenShift users to login to Jenkins via an OAuth flow. The plugin syncs and maps the users permissions in OpenShift to Jenkins permissions. The plugin can be configured to only authorise users who have read or edit access to the same namespace as where Jenkins is deployed.
+
+#### Option 3: Add a reverse proxy
+
+This solution can be used in the following situations:
+
+* The Service has no authentication e.g. Prometheus
+* The Service has authentiation, but supports a 'http reverse proxy' e.g. [Grafana](http://docs.grafana.org/installation/configuration/#auth-proxy)
+
+An example of both these approaches was progressed in https://github.com/aerogearcatalog/prometheus-apb/blob/51cdd00f1c562537c72cbeb298a0c3fb00d321b8/roles/provision-prometheus-apb/tasks/main.yml
+This example used [OAuth Proxy](https://github.com/openshift/oauth-proxy). The proxy presents the User with an OAuth flow for OpenShift users to login. The proxy can be configured to protect certain (or all) endpoints in the upstream service. It can also be configured to do a SubjectAccessReview (SAR) to ensure the user has certain permissions e.g. can a user 'view' 'Services' in the current namespace.
--- a/auth/developer-single-sign-on-across-mobile-services.md
+++ b/auth/developer-single-sign-on-across-mobile-services.md
@@ -9,7 +9,7 @@ This document outlines a proposal for single sign-on across Mobile Services & Mo
 
 ## Problem Description
 
-As a developer builds up their Project by provisioning different services, they will likely need to setup or configure ascpects of those services. For example, if a developer provisions the Push Notifications service, they will need to configure Push credentials for Mobile Apps. Some services may also provide additional info or dasboards. Accessing these dashboards & performing setup or configuration will likely require signing into those services via some auth mechanism.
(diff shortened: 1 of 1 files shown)


auth/developer-single-sign-on-across-mobile-services.md
```diff
@@ -0,0 +1,59 @@
+# Abstract
+
+This document outlines a proposal for Single Sign On across Mobile Services & Mobile Enabled Services provisioned on OpenShift. This proposal is *not* intended as a feature for end users of Mobile Apps. It *is* intended as a development time feature for Mobile developers.
```
wtrocki: I know it's very small stuff but in most of the cases I have seen name`Single sign-on` 
For example: https://en.wikipedia.org/wiki/Single_sign-on

wtrocki: `This proposal is *not* intended as a feature for end users of Mobile Apps. It *is* intended as a development time feature for Mobile developers.` 

Any example will be good here to make sure that people will understand it.

auth/developer-single-sign-on-across-mobile-services.md
```diff
@@ -0,0 +1,59 @@
...
+
+However, with Keycloak this method is limited to only automatically setting up the initial permissions of *all* users who can successfully authenticate. There is no permissions mapping between OpenShift roles & Keycloak administration roles. Additionally, *any* OpenShift user will be able to sign in to Keycloak. Manual configuration is required to only allow certain users access and limit their roles.
+
+#### Option 2: Use a plugin for OAuth
+
+If a Service allows plugins, it may be possible to use a plugin for auth. For example, the [Jenkins OpenShift Login Plugin](https://github.com/openshift/jenkins-openshift-login-plugin#openshift-login) allows OpenShift users to login to Jenkins via an OAuth flow. The plugin syncs and maps the users permissions in OpenShift to Jenkins permissions. The plugin can be configured to only authorise users who have read or edit access to the same namespace as where Jenkins is deployed.
+
+#### Option 3: Add a reverse proxy
```
wtrocki: This will be the most flexible, language independent option IMHO. 
When testing this now each separate oath proxy will require login. 

Is this assuming that we will use single oauth proxy per project (that may be tricky) or there is a way to do SSO on multiple different ones? Excited to see how you will see this done.

david-martin:
@TommyJ1994 would you mind reviewing this too?
I can't add you as a Reviewer in the top right, maybe you need to be added to a team

auth/developer-single-sign-on-across-mobile-services.md
david-martin: My initial thinking is a different oauth proxy per service that needs it.
OAuth Proxy does have support for multiple upstreams https://github.com/openshift/oauth-proxy#upstream-configuration, but I haven't tried it.
Also, I don't know how feasible that is technically if different services, provisioned at different times, need to use the same proxy.

david-martin: I don't think this could be the solution for all services. Using the reverse proxy assumes you don't have auth in the upstream service, or your upstream service has some way of distinguishing users based on a proxied header. Identifying the user in the upstream service, so roles/permissions can be mapped accordingly could be important for that service.

tomjackman:
@david-martin Keycloak has its own [Security Proxy](http://www.keycloak.org/docs/latest/server_installation/index.html#_proxy) than can be used to front app's/services where a keycloak or OIDC library can't be placed. I'm not 100% sure if it fits this use case as I havn't looked into before.

The first approach is probably the most traditional (basis for identity brokering) and ones that Keycloak supports best. Even though you can apply default group roles, it could still be a bit more finer grained. They do support LDAP role mapping, but It might be worth asking the Keycloak team if its on their roadmap to also include OAuth/OIDC role mappings on import.

Its not the easiest decision to make, as the flow is a bit different to the norm:
- Identity Provider --> Allows Login using another App/Service (Openshift)
- App/Service (Openshift) --> Authenticates a user and allows them to control/carry out actions on the Identity Provider from "outside" the Identity space. 

I'm guessing their is no way to flip it around and make Openshift users authenticate to the Identity Manager (Keycloak) instead, rather than Openshift being the source of the identities? Let me know if some if this is out of context. I'm trying to not confuse it with some previous conversations we had.

wtrocki:
@TommyJ1994 Have you used/tested that: https://issues.jboss.org/browse/KEYCLOAK-4528 ?

auth/developer-single-sign-on-across-mobile-services.md
```diff
@@ -0,0 +1,59 @@
...
+This document outlines a proposal for single sign-on across Mobile Services & Mobile Enabled Services provisioned on OpenShift. This proposal is *not* intended as a feature for end users of Mobile Apps. It *is* intended as a development time feature for Mobile developers.
+
+## Terms
+
+* Mobile Services - Services that provide mobile specific functionality for Mobile Apps at development time or runtime e.g. Push Notifications, Data Sync, CI/CD
+* Mobile Enabled Services - Services that provide functionality for Mobile solutions at development time or runtime, but functionality is not specific to Mobile e.g. Keycloak, Metrics.
+
+## Problem Description
+
+As a developer builds up their Project by provisioning different services, they will likely need to setup or configure ascpects of those services. For example, if a developer provisions the Push Notifications service, they will need to configure Push credentials for Mobile Apps. Some services may also provide additional info or dasboards. Accessing these dashboards & performing setup or configuration will likely require signing into those services via some auth mechanism.
```
wtrocki: Typo `dasboards`

david-martin:
@TommyJ1994 The Security Proxy sounds interesting.
Is it able to do custom permission checks based on user info in OpenShift?
I don't think the OpenShift user token has any info on roles, but there is a user info endpoint that will give role info. Based on that, it would be nice to check if the user has edit/view access to the project or some other resource.

I was chatting with @wei-lee where he brought up the keycloak proxy approach.
My understanding with that suggestion is that keycloak would be required for all services.
For example, provisioning fh-sync-server would also provision keycloak. Same for build farm, metrics etc...
When chatting with @maleck13 & @johnfriz about this before, we shyed away from it as it meant keycloak was a dependency for everything before even provisioning a service. So if you wanted to do anything mobile, straight away you needed keycloak in your project.
That isn't something we wanted to impose on developers, so hooking into OpenShift auth was something we discussed as a way of doing single sign-on for service administration (hence this proposal).
Keycloak could definitley be the single sign-on approach for Mobile App users though

@TommyJ1994 @wei-lee is this something you'd like to propose again and discuss wider?

auth/developer-single-sign-on-across-mobile-services.md
wei-lee: For this approach, I think another option worth discussing is to use Keycloak as the single auth service provider, and use the "Identity broker feature" provided by KC and the [Keycloak-proxy](https://github.com/gambol99/keycloak-proxy) to perform user authentication against OpenShift 3.

There are a few benefits of doing this way:
1. It is possible for developers to add more identity providers through Keycloak. I can certainly see a use case where the developer may want more users to access the service, but they may not necessarily be OpenShift users (e.g. business users want to see some data from a analytics service). This option will make it easy to do by the developer.
2. All the services only needs to talk to Keycloak for authentication, instead of using OpenShift to authenticate admin users, and then KC to authentication mobile users. 
3. Provide a good example of how to perform user authentication for developer-defined services. We have said it before that when comes to user authentication, Keycloak should be the one and only provider that we support. I think we should stick to that for our own services as well.

However, the biggest challenges of doing this way is that all the other services will become dependent on Keycloak, so we will probably have to provision Keycloak as part of the provision process for other services, and also configure keycloak to add Openshift as an identity provider. However, the way I think is that allow Openshift users to access the services *IS* a user authentication requirement, and in this case, it makes sense to provision Keycloak (a.k.a Identity management service) together.

@david-martin @maleck13 @matzew @johnfriz would like to hear your thoughts on this.

wei-lee:
@david-martin @TommyJ1994 please see my comment above.

auth/developer-single-sign-on-across-mobile-services.md
wtrocki: @wei-lee Propositions sounds good, but having keycloak from beginning sounds like big deal for online first aproach. 

What about having keycloak as optional service instead of provisioning that from the beginning? For example if I want to demo single service in cluster with restricted resources I will be able to still demo that without auth. 

Do you think it's possible to provision keycloak service later, at any stage of development?

wei-lee: > but having keycloak from beginning sounds like big deal for online first aproach

Given that KC has support for multi-tenancy, one option we can consider is to have a KC instance running in a system name-space (or a namespace created by the mobile add-on). When a developer provisions a new service,  he/she can choose if the default KC instance should be used, or a new one should be created.

> What about having keycloak as optional service instead of provisioning that from the beginning? For example if I want to demo single service in cluster with restricted resources I will be able to still demo that without auth.

It is possible. In that case, we can just simply present the service without any user authentication. If they want to authentication, they can add KC service later on.

wtrocki: > It is possible. In that case, we can just simply present the service without any user authentication. If they want to authentication, they can add KC service later on.

+100

auth/developer-single-sign-on-across-mobile-services.md
maleck13: type ascpects aspects

auth/developer-single-sign-on-across-mobile-services.md
```diff
@@ -0,0 +1,59 @@
...
+
+For cases where no auth mechanism is provided with a service, a reverse proxy approach that uses OpenShift OAuth, as shown in Option 3 below, can be used.
+
+### Provider Integration Options
+
+Choosing how to integrate a Service with OpenShift OAuth will depend on what the Service is capable of already, and how extensible it is. The below options show practical examples of how this integration was done. There will likely be tweaks and other options required depending on the service. These options serve as references for how it could be done.
+
+#### Option 1: Configure OAuth in the Service
+
+If a Service has the ability to an OAuth provider, use it. For example, in Keycloak it is possible to use 'Identity Brokering' to defer to OpenShift Auth for who can administer that Keycloak instance. See this [Keycloak Identity Brokering Blog Post](https://developers.redhat.com/blog/2017/12/06/keycloak-identity-brokering-openshift/) for details.
```
maleck13: replace: If a Service has the ability to an OAuth provider,
with: If a Service has the ability to integrate with an OAuth provider,

auth/developer-single-sign-on-across-mobile-services.md
```diff
@@ -0,0 +1,59 @@
...
+
+However, with Keycloak this method is limited to only automatically setting up the initial permissions of *all* users who can successfully authenticate. There is no permissions mapping between OpenShift roles & Keycloak administration roles. Additionally, *any* OpenShift user will be able to sign in to Keycloak. Manual configuration is required to only allow certain users access and limit their roles.
```
maleck13: is the manual configuration on a per user level or is it something that we could look to do from the APB?

david-martin: Per user level.
Working with @wei-lee to figure out if this could be automated. Will be chatting with keycloak dev members too

auth/developer-single-sign-on-across-mobile-services.md
maleck13: > Given that KC has support for multi-tenancy, one option we can consider is to have a KC instance running in a system name-space (or a namespace created by the mobile add-on). When a developer provisions a new service, he/she can choose if the default KC instance should be used, or a new one should be created.


When you say system space it seems to imply a managed service. Not sure this would work as the question of who would manage it remains

wei-lee: @maleck13 can it be the people who manages openshift manage it? It will be treated like any other system managed projects.

pb82: @wtrocki @wei-lee if we give users the option to add KC as a service later (and have no auth until then) doesn't that also mean that we not only have to add KC but also keycloak-proxy to all services to enable authentication for them?

wei-lee: @pb82 yes, if the service doesn't support KC, or doesn't provide authentication function at all. 

For example, for sync & push, they already support KC so no need for the keycloak-proxy. For services like prometheus,  it doesn't support authentication so it's best to add keycloak-proxy.

david-martin: @wei-lee based on offline discussions, are you happy if we add an Option 4 for a keycloak-proxy approach, and merge the proposal then (after re-review)?

I think having keycloak as the **only** auth provider in their namespace (for both administration & app users) can be an option for developers to migrate to, rather than it being an 'out of the box' option. I think the overhead for requiring a keycloak for every service (even if its shared/managed), both resource overhead (for developers) and maintenance/support overhead (for us) could be quite high. 
I think minimizing the barrier for developers to try out mobile services is more important in this case.

wei-lee: @david-martin I don't think it's a overhead for us, because we have to maintain/support it anyway as part of the authentication service offering. But I do agree that it is an overhead for developers to start with if they only need admin access to a service. 

So yeah, I am happy to go ahead as you suggested.

david-martin: Thanks @wei-lee.
re: overhead for us, I was thinking the extra overhead would be us adding keycloak auth support/maintenance for services that wont be accessed by app users or wont require auth when accessed by app users e.g. Jenkins/Build Farm, Metrics.

(Metrics will likely need some sort of api key support for app users, but definitely not require users to login)

david-martin:
@wei-lee Added option 4 https://github.com/aerogear/proposals/pull/8/commits/cd2e7b3b00a4381fb6b0351d4c39b85361816aa0
I'll hold off on merging until your final approval.

wei-lee:
@david-martin The options looks good. To complete the proposal, I think it will be helpful to add a section to describe the options that will be taken for the services that we already know.

david-martin:
@wei-lee added a section (and shorter Option titles) https://github.com/aerogear/proposals/pull/8/commits/4e44bb129377786742ab417d0540eb1aadc52293

wei-lee (approved):

PR #9, opened 2018-01-09
Android sdk proposal

secondsun:
Here is a new proposal for the Android SDK.


secondsun:
I apologize in advance for typos and such.  It is a late day commit :)

android-sdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+                  appKey=”xyz”)
+public interface UploadService {
+   @ServiceMethod
+   public RequestHandle upload(Picture mypicture);
+}
+```
+In this world, the `@MobileService` annotation would signal to the core SDK that this interface should be implemented by our libraries and it requires a request to include the Keycloak, Sync, and Storage.
+The code could then be invoked by the developer as follows : 
+```
+RequestHandle handle = org.aerogear.android.core.MobileService.from(UploadService.class).upload(myPicture);
```
wei-lee: I am not entirely sure how this works. More specifically, I am not sure how `MobileService` will know what methods to invoke in the annotated classes. 

For example, the `KeycloakService` may have multiple methods, e.g. login, logout, isTokenExpired, addTokenHeader etc. In this case, the `MobileService` should invoke the `addTokenHeader` method. But how does it know which one to invoke?

android-sdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Proposed solution
+
+### Core SDK
+
+The Core SDK will be the package that all service modules inherit from.  It will be responsible for managing network requests, logging, error handling and resolution, loading configuration for mobile-core services, and other cross cutting concerns.  Modules will interface with the core SDK by implementing the necessary interfaces provided and annotating the implementations so that the Core SDK will be able to correctly manage them as well.  The Core SDK will also allow some of its internals to be exposed through interfaces for purposes such as testing, bespoke integrations, and management of additional resources.
```
wei-lee: I think the core another function that should be provided by the core module is threading. The core module should provide the same threading pattern for all the services (e.g. using RxAndroid or Bolts)

android-sdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+### Primary Goal
+ Provide a starting point for development of the Android SDKs based on the general architecture described in https://github.com/aerogear/proposals/blob/1b79c6323997a0fa4e9794bb2b91145702cf5a48/sdk/overview.md as well as define for Android what tools we will build (ie build plugins, IDE tooling, scripts, etc), the relationship between the "Core" SDK and "Service" SDKs, release and distribution processes, and processes for ensuring code quality (dependency management, static analysis, CI/CD, etc)
+
+## Terms
+### Core SDK
+The base library that provides the mechanisms for setting up mobile-core services.  It will include interfaces for Requests, Responses, Services, Modules, and mechanisms for how they become composed.
+### Service Module
+A Service Module connects to a OpenShift service that provides some function.  Examples of services are sync, keycloak, push, rest with RHOAR runtimes, etc
+### Application Module
+This is the code which is written by an application developer to bridge our modules and their application. (for example, application specific business logic)
```
wtrocki: Can we assume that modules are internal part of our sdk and users will just import that?
We should handle scenarios when sdk can be added to the existing application.

android-sdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Proposed solution
+
+### Core SDK
+
+The Core SDK will be the package that all service modules inherit from.  It will be responsible for managing network requests, logging, error handling and resolution, loading configuration for mobile-core services, and other cross cutting concerns.  Modules will interface with the core SDK by implementing the necessary interfaces provided and annotating the implementations so that the Core SDK will be able to correctly manage them as well.  The Core SDK will also allow some of its internals to be exposed through interfaces for purposes such as testing, bespoke integrations, and management of additional resources.
+
+The Core SDK will provide a request lifecycle that the Psuedocode section will demonstrate.
+
+### Service Modules
```
wtrocki: It's good to explain what module means. I understand it as separate maven dependency that you add in gradle config. Is that right?

android-sdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+Service modules will each represent a singular service that is supportd by mobile-core.  We will use, where available and practical, existing libraries for services and in those cases the modules will provide the necessary implementation of Core APIs to make them compatable with the Core SDK lifecycles and features.  The first Services we are looking at integrations will be Keycloak, fh-sync-service, and a RESTful API service module.
```
wtrocki: `RESTful API service module.` seems good use case for testing this, without getting into complexity of service. We will also have stats service prototype.

android-sdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+There will be, in OpenShift, a Mobile application, a sync server, a keycloak server, and a storage server.  An implementation of this interface could load a keycloak token from the device, validate the token, start a new OkHttpRequest object, Upload the file etc etc.  What I am getting at is that this is a very laborious process.  However, let’s annotate this call a bit.
+```
+@MobileService(requires = {KeycloakService.class, SyncService.class, StorageService.class},
+                  appKey=”xyz”)
+public interface UploadService {
+   @ServiceMethod
+   public RequestHandle upload(Picture mypicture);
+}
+```
+In this world, the `@MobileService` annotation would signal to the core SDK that this interface should be implemented by our libraries and it requires a request to include the Keycloak, Sync, and Storage.
```
wtrocki: Core SDK will scan application to check for @MobileService annotation?

android-sdk/README.md
wtrocki: I think it may need more code.  We may do couple different approaches for SDK (as prototypes) and pick best patterns.

android-sdk/README.md
secondsun: I'm not sure what you mean by "internal".  

The way I see my proposal working is that there is the app and code that calls our modules.  Some of that code may use the SDK to register itself as a module (ie if the developer has a custom piece of code that they want managed by core).  This is what an application module would be.

Service modules are provided by us and integrated with the core sdk already.  They aren't supposed to be used directly by the developer, but can expose some some apis if it is appropriate.

It is like in firebase how there are multiple modules you import in build.gradle, but they are exposed via a few well known APIs.

android-sdk/README.md
secondsun: +1.

android-sdk/README.md
secondsun: Yes.  I'll make a note to add a example of how it would look in code.

android-sdk/README.md
secondsun: That's what I was thinking.  It will also be the bulk of the custom code that is written in mobile-core.

In the space future we may have websocket services, push based vert.x event bridges, etc.  But for now, we REST :)

android-sdk/README.md
secondsun: "Yes".

If we make a plugin, the scan can be done at compile time and we can generate the instanciation code then.  (Dagger 2 style).  Without the plugin we can scan at startup/lazy loading (Dagger classic style)

android-sdk/README.md
wtrocki: @secondsun I did not read that properly. After your explanation it's clear for me how this will work.

android-sdk/README.md
secondsun: Dependency injection, either with compile time or run time classpath scanning will bridge annotated services and the MobileCore endpoints.

Something we need to work out is the idea of a request and a response lifecycle.  In AeroGear Android Pipe this was done using PipeModules which defined a few methods that are called during a request/response.  Pipe also had mechanisms for controlling how requests and responses were written to and parsed from HTTP services.

Links : 
--Modules
Pipe Module : https://github.com/aerogear/aerogear-android-pipe/blob/master/library/src/main/java/org/jboss/aerogear/android/pipe/module/PipeModule.java
Pipe Module Implementation (OAuth2) : https://github.com/aerogear/aerogear-android-authz/blob/master/library/src/main/java/org/jboss/aerogear/android/authorization/oauth2/OAuth2AuthzModule.java

-- Request/Response
RequestBuilder : https://github.com/aerogear/aerogear-android-pipe/blob/master/library/src/main/java/org/jboss/aerogear/android/pipe/RequestBuilder.java
Response Parser : https://github.com/aerogear/aerogear-android-pipe/blob/master/library/src/main/java/org/jboss/aerogear/android/pipe/ResponseParser.java

Example app showing all of this working together : https://github.com/aerogear/aerogear-android-cookbook/blob/master/ShootAndShare/app/src/main/java/org/jboss/aerogear/android/cookbook/shootandshare/util/GooglePlusHelper.java

danielpassos: This code means I can do what the anotation process will do by my self?

secondsun: Yes.  The goal is to make annotations be the golden scenario and work for most use cases.

android-sdk/README.md
secondsun: @wei-lee WDYT? https://github.com/googlesamples/android-architecture/blob/todo-mvp/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/util/AppExecutors.java

The thing I like about this is that the utility class provides threads for certain tasks.  Splitting Disk IO and networking into different threads lets the application do things like show a cached resource, dispatch a call for a remote resource in the cache, and then update the cache while the cache is being displayed to the user.  Since there is only one thread per use we don't hit any weird issues where the cache is updated in the middle of a read giving developers nightmares.

android-sdk/README.md
wei-lee: @secondsun it's still not clear to me how the `MobileService` implementation can find the right method to invoke. Are you saying that the methods in `KeycloakService` and `SnycService` will be annotated? Do you have a little bit more examples to explain how this works?

android-sdk/README.md
wei-lee: I think I quite like it. It's simple, powerful, easy to use and doesn't require any dependencies. So +1 from me.

What about returning the results back to to caller? Are we thinking just let the caller blocked and wait for the tasks to be completed, or callbacks, or promises?

secondsun: I think we will also want to include testing structures in the core sdk, or in a core-testing sdk that is, archetecturally, at the same level as the core sdk.

secondsun: @wei-lee I'll show you how it works with the caller in a bit :)

android-sdk/README.md
secondsun: @wei-lee Right now my thought is either the service and its methods will be an annotated POJO, or the service will be annotated at the class level and implement a interface for the methods.

android-sdk/README.md
secondsun: @wei-lee Updated with a threading example

johnfriz:
Can we add some detail on the structure / architecture of the Android SDK - e.g. what is required to add a new service. 

Can we also add some info on how docs will be handled - Android API Docs and how they tie into the higher level service docs.

secondsun:
@johnfriz Are you asking for technical docs (JavaDocs) or friendly docs (tutorials etc)?
Is there a proposal for service docs I can reference?

wei-lee:
Based on the discuss we had today, I think the following content can be added to this proposal:

1. a high-level simple diagram to show the main modules that will be provided by MobileCore. The relationship between MobileCore and a service module (for example, IDM)
2. For a service module, what interface it should implement/extend? What's the main functions/responsibility of the service module?
3. For a service module, what's a typical constructor looks like?
4. How to initialise/bootstrap the MobileCore? What the initialise/bootstrap function will be doing?
5. How can developer get the implementation for a particular service? e.g. how can developer get the IdmService?
6. How one service module can use another service module? e.g. how the sync service can use the IdmService to add token to the request?

secondsun:
@wei-lee I couldn't get the class diagrams finished, but if you look here : https://github.com/secondsun/aerogear-android-sdk/tree/PoC I've stubbed the new information I've put into the proposal.

android-sdk/README.md
```diff
@@ -0,0 +1,216 @@
...
+The Mobile Core SDK will provide Java Executors that wrap several common blocking tasks.  These executors will be available to any module or part of the application which may need to schedule diskIO work, netowrking work, or work to happen on the main thread.
+
+```
+AppExecutors.networkThread.execute(() -> {
+  var result = someRequest.call(); //Perform a HTTP operation off the main thread
+  AppExecutors.mainThread.execute(()->{updateDisplay(result);});//Update the Android UI on the main thread.
+ });
+```
+
+#### Deprecated as we are moving away from annotations for the initial version
```
wei-lee: if this is deprecated, I say let's remove this section for now.

android-sdk/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+Service modules will each represent a singular service that is supportd by mobile-core.  We will use, where available and practical, existing libraries for services and in those cases the modules will provide the necessary implementation of Core APIs to make them compatable with the Core SDK lifecycles and features.  The first Services we are looking at integrations will be Keycloak, fh-sync-service, and a RESTful API service module.
+
+### Implementation notes
+
+Public APIs should always expose native Java Objects and not stringified JSON or JSON Objects.
+
+### Use cases and psuedo code examples
+
+#### Staring up Mobile Core
```
wei-lee: typo: `Starting`

wei-lee:
@secondsun let's added the diagram I created and clean up the proposal a little bit. Then I am happy to accept the proposal.

android-sdk/README.md
```diff
@@ -0,0 +1,216 @@
...
+      Reuqest request = networkService.newRequest();
+      request.addSecurityHeaders(keycloakService.getHeaders());//pseudocode method, don't expect this exactly.
+      request.setData(this.buildrequestBody());
+      Response response = networkService.execute(request);
+      response.onError(/*Error handling lambda*/);
+      response.onSuccess(/*Success handling lambda*/);
+    }
+  }
+
+}
```
wei-lee, line 133: how this will work if the auth service is not enabled, or if developers want to use a different authentication service that is not provided by us with sync?

android-sdk/README.md
```diff
@@ -0,0 +1,186 @@
+# Android Core SDK + Auth, REST, and Sync Service modules Proposal 
+This document outlines the features, architecture choices, tooling needs, and processes for developing the Android SDKs for communicating with services connected by Mobile-Core.  It is hoped that This can also be a model for other platforms as we move forward.
+
+## Goals
+
+### Primary Goal
+ Provide a starting point for development of the Android SDKs based on the general architecture described in https://github.com/aerogear/proposals/blob/1b79c6323997a0fa4e9794bb2b91145702cf5a48/sdk/overview.md as well as define for Android what tools we will build (ie build plugins, IDE tooling, scripts, etc), the relationship between the "Core" SDK and "Service" SDKs, release and distribution processes, and processes for ensuring code quality (dependency management, static analysis, CI/CD, etc)
```
david-martin: This might read better as smaller bullet points e.g.

* Provide a starting point for development of the Android SDKs based on the general architecture described in https://github.com/aerogear/proposals/blob/1b79c6323997a0fa4e9794bb2b91145702cf5a48/sdk/overview.md
* define for Android what tools we will build (ie build plugins, IDE tooling, scripts, etc)
* the relationship between the "Core" SDK and "Service" SDKs
* release and distribution processes
* processes for ensuring code quality (dependency management, static analysis, CI/CD, etc)

android-sdk/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Terms
+### Core SDK
+The base library that provides the mechanisms for setting up mobile-core services.  It will include interfaces for Requests, Responses, Services, Modules, and mechanisms for how they become composed.
+### Service Module
+A Service Module connects to a OpenShift service that provides some function.  Examples of services are sync, keycloak, push, rest with RHOAR runtimes, etc
+### Application Module
+This is the code which is written by an application developer to bridge our modules and their application. (for example, application specific business logic)
```
david-martin: I find this concept of an application module, and how it integrates with/uses the Core & Service modules difficult to grasp.
Do developers write application modules?
Is it just custom code in their app?
Do they need to follow some hard patterns or interfaces to work with the Core & Service Modules?
Do they have to divide their app into applicaiton modules, and register them all with the Core SDK?

android-sdk/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+The Core SDK will provide a request lifecycle that the Psuedocode section will demonstrate.
+
+### Service Modules
+
+Service modules will each represent a singular service that is supportd by mobile-core.  We will use, where available and practical, existing libraries for services and in those cases the modules will provide the necessary implementation of Core APIs to make them compatable with the Core SDK lifecycles and features.  The first Services we are looking at integrations will be Keycloak, fh-sync-service, and a RESTful API service module.
+
+### Implementation notes
+
+Public APIs should always expose native Java Objects and not stringified JSON or JSON Objects.
```
david-martin: :+1: This was a big problem in many parts of millicore, leading to validation issues and NPEs

android-sdk/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Use cases and psuedo code examples
+
+#### Staring up Mobile Core
+MobileCore is the "main" class for the SDK of the Core SDK.  Users instanciate it by placing a client configuration file in their assets directory, configuring MobileCore.Builder, and then calling its build method.
+
+```
```
david-martin: if we add `java` after the backticks, we'd get syntax colouring.

android-sdk/README.md
```diff
@@ -0,0 +1,186 @@
...
+    public static void registerServiceModule(String type, Class<? extends ServiceModule> moduleClass, String... dependsOn) {
+        serviceTypeMap.put(type, moduleClass);
+        for (String dependency : dependsOn) {
+            dependencyMap.put(type, dependency);
+        }
+    }
+
+}
+```
+** Discussion Point ** I'm not a fan of static methods during testing.  Perhaps there should be a method in the ServiceModule interface which returns a list of services that the service depends on.  MobileCore could then read that list by invoking that method as part of bootstrapping.
```
david-martin: I'm also not a fan of static methods for testing reasons. It allows for state pollution between test runs. However, there are ways around it, like a static reset method that must be run after each test. 
If I was progressing this work, I'd probably start out with static and see if there's a reasonable way to move away from it that doesn't overburden the developer

android-sdk/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### Referencing a Service
+An instance of the core bassed during bootstrap can be used to reference other services.  MobileCore is responsible for ensuring that ServiceModule.bootstrap is not called until all declared dependencies are bootstrapped.
```
david-martin: (I'm assuming you're using MobileCore and Core SDK interchangeably here)
How will MobileCore know when this is? 
Does the developer have to do that at a point in their code where all services have been declared/registered with the Core SDK?

android-sdk/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### Using Core Networking
+For now, networking will be a service exposed by a http service module.  The definition of this service will be a different proposal.
+
+#### Proposal diagram
+![Core Service and Module diagram](./img/diagram.png)
+
+
+### Misc. Bits
+The following topics are not required
```
david-martin: I'm not understanding this line in this context.
Do you mean you don't think any of these should be done?
Or they should be done, but not part of this proposal?

david-martin: Based on the primary goal, the sub sections here should be given more importance than 'not required'

* as well as define for Android what tools we will build (ie build plugins, IDE tooling, scripts, etc)
* release and distribution processes
* and processes for ensuring code quality (dependency management, static analysis, CI/CD, etc)

If they aren't required, should they be part of the primary goal then?

david-martin (changes requested):

android-sdk/README.md
secondsun: @david-martin A application module is a service module that is written by an application developer.  It is a way for us to distinguish in discussion code that we provide vs code that a user is expected to write.

An application module will be registered with the service repository, bootstrapped by core during build, and will be how modules expose themselves to each other.

android-sdk/README.md
secondsun: +1

android-sdk/README.md
secondsun: MobileCore is a class. Core SDK is the SDK.  Any places where it is ambiguous should be clarified.

MobileCore will be resposible for bootstrapping methods during instantiation/initialization. (see the https://github.com/secondsun/aerogear-android-sdk/blob/87b4ff629b6f3c47040d2e322277e7cee3bcb76f/core/src/test/java/org/aerogear/mobile/core/MobileCoreParserTest.java#L25)

Services have to be declared in the service registry before then.

android-sdk/README.md
secondsun: They should be done, but might not be part of this proposal.

The proposal has narrowed its scope in the past week or so from a "all powerful android sdk proposal" to "Core SDK setup and module system" which is much more manageable.  This language may need to be reworked in that light.

android-sdk/README.md
secondsun: So for activities defined in this proposal

sync (dependsOn) keycloak (depends on http)

right now we don't have a mechanism for querying the state of KeyCloak in a way that is abstract from sync, so sync will need to do it on its own.  Maybe as part of IDM we can figure out a generic "authentication" API that sync can consume and anyone can implement?

android-sdk/README.md
secondsun: fixed

android-sdk/README.md
david-martin: Thanks @secondsun. Are application modules necessary for a developer to use our SDK(s)?
Do they have to register things with the service repository other than the Service SDKs/modules they intend to use in the app code?

android-sdk/README.md
david-martin: So is the instantiation/initialization something the developer triggers programatically in their app code? e.g. `mobileCore.initialize()`
And they have to make sure they've declared all services with the registry before they call that method?

android-sdk/README.md
david-martin: OK. It may be enough to remove them from the Primary Goals (or move them to a Non Goals section if you still want to draw some attention to them).

david-martin:
> Based on the discuss we had today, I think the following content can be added to this proposal:
> 
> a high-level simple diagram to show the main modules that will be provided by MobileCore. The relationship between MobileCore and a service module (for example, IDM)
> For a service module, what interface it should implement/extend? What's the main functions/responsibility of the service module?
> For a service module, what's a typical constructor looks like?
> How to initialise/bootstrap the MobileCore? What the initialise/bootstrap function will be doing?

@wei-lee I agree that having all of these in the proposal is desirable. I think most of that is in the proposal now

> How can developer get the implementation for a particular service? e.g. how can developer get the IdmService?
> How one service module can use another service module? e.g. how the sync service can use the IdmService to add token to the request?

These ones could take some time to get right and may only 'come out in the wash'. I wouldn't see it as a problem if the proposal was merged without these fully figured out.
That said, I like the idea of the Core providing the HTTP bit, and being able to supplement requests/responses/errors. That sounds like a core part of solving this.

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+    private static final Map<String, Class<? extends ServiceModule>> serviceTypeMap = new HashMap<>();
+    private static final Map<String, String> dependencyMap = new HashMap<>();
+    /**
+     * Register the looking for a service module to a class.
+     *
+     * @param type the type of module to use
+     * @param moduleClass the class that implements a service module type
+     * @param dependsOn the other serviceTypes this Service Depends on.
+     */
+    public static void registerServiceModule(String type, Class<? extends ServiceModule> moduleClass, String... dependsOn) {
```
wtrocki: After conversation with @secondsun 
This requires spike how dependencies are needed.
Current use case is that for example: sync has dependency on keycloak. 

We should investigate this case after Networking layer is done.

android-sdk/README.md
wtrocki: 🗣  Element to cover

My comment above tackles the same issue. 
We may just skip to mentioning application modules/plugins initially or explain it.
Needs wider agreement.

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+MobileCore.Builder builder = new MobileCore.Builder(application);
+MobileCore core = builder.build();
+```
+
+During the build method, MobileCore will scan a registry of services and configure them by calling the method `bootstrap` defined in the ServiceModule interface.  MobileCore will pass to bootstrap the core instance and the section of mobile-core.json that corresponds to that service.
+
+#### Writing a Service Module
+A service interface will be responsible for building requests, parsing responses, handling errors.  Services may depend on each other and the method of service composition is currently under evaluation. However, services will have a resolution order where a request is passed along a chain until it is ready to be executed.  Mobile Core will provide the services `threading`, `logging`, and `networking`.  ServiceModules will implement the service module class and provide a default no argument constructor.  
+
+Example KeyCloakService
```
wtrocki: [Oftopic] Keycloak service may not be the best first service to create as it may be considered an wrapper for networking layer. It will be cool to see sync service as we already have it extracted and we can simply import that into first production ready service. 

IMHO Having sync will allow us to do first service that will work with full end to end integration with openshift. We could then introduce security and keycloak by spiking two options:

- Keycloak and security as part of the core
- Keycloak as separate service

🗣  Thing to talk about.

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+        this.subjectTokenType = config.getProperty("subject_token_type");
+        this.requestedTokenType = config.getProperty("requested_token_type");
+        this.realm = config.getProperty("realm");
+        this.core = core;
+    }
+}
+```
+
+#### Registering a Service Module and the ServiceModuleRegistry
+The Core SDK will provide a service registry that modules can register themselves with as well as register their dependencies.  During build MobileCore will call bootstrap on services in the order of those with the least dependencies to those with the most dependencies.
```
wtrocki: After conversation with @secondsun around dependency model. 
This requires spike if dependencies are needed and what problem this dependency mechanism will resolve.
Current use case is that for example: sync has dependency on keycloak.

IMHO if we can avoid this by just allowing core networking layer to be extended by keycloak service.
Best for services to be independent and do not relay on each other. I might misunderstood something here.

We should investigate this case after Networking layer is done so we will have more real life examples how tis can be implemented.

🗣  Let's talk about it.

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+Example registration of KeyCloakService
+```
+public class KeyCloakService implements ServiceModule {
+  static {
+    ServiceModuleRegistry.registerServiceModule("keycloak", KeyCloakService.class, "http");
```
wtrocki: ServiceModuleRegistry has static method (as mentioned above we should avoid that) so we can make it singleton

android-sdk/README.md
wtrocki: @david-martin Has really good point. I think that proposal should first define top level user api. For example:

1. Add core and serviceA dependency to gradle
2. Create Core instance by.. etc.

This will help us to see how users will use it and we can then describe internal parts

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+An instance of the core bassed during bootstrap can be used to reference other services.  MobileCore is responsible for ensuring that ServiceModule.bootstrap is not called until all declared dependencies are bootstrapped.
+
+SyncService.java
+```
+public class SyncService implements ServiceModule {
+
+  KeyCloakService keycloakService;
+  /*snip*/
+  public void bootstrap(MobileCore core, ServiceConfiguration config) {
+     this.keycloakService = core.getService("keycloak", KeyCloakService.class);
```
wtrocki, line 117: Do we se any use case now when for example Sync service will need to internally require push etc.?
Is this a target for this code?

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+     this.networkService = core.getService("network", NetworkService.class);
```
wtrocki, line 118: What `getService` is getting as parameters? 
Why users will use strings and class to get service? 

It seems that we mixing 2 different things - internal core functionalities like networking with separate service SDK like sync. 

I think it will be just better for core to have explicit list of internal functionalities like networking by:

```java
// Usage in any module like sync
core..getNetworkBuilder().url(url).get(...).build()

// Extend core in keycloak service - if keycloak service is not part of the core.
core.getNetwork().setInterceptor(keycloakInterceptor)
```

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+     /*snip*/
+  }
+
+  public void performSync() {
+    if (keycloakService.isLoggedIn()) {
+      Reuqest request = networkService.newRequest();
+      request.addSecurityHeaders(keycloakService.getHeaders());//pseudocode method, don't expect this exactly.
```
wtrocki: -1 for this aproach. If I will have ability to change single element in proposal it will be this statement.
Imposing hard dependencies on the services itself (managed by us) it's not flexible aproach and it may be hard to maintain. It has numerous implications that server layer needs to handle. 

Happy to explain that carefully 🗣 

TL;DR - That's imposing hard dependency on sync service and it's less flexible.

android-sdk/README.md
wtrocki: From my point of view I would want to know how we going to enable security on the server side. 
If keycloak is being provisioned does that mean that all services will automatically become keycloak protected using some proxy or direct implementation? I think that format for `mobile-services.json` may be not flexibile enough - keycloak now is now as separate service - while the way it should be done is that keycloak config should be part of the each service that is protected by that. 

By this we will be able to:
- Determine if service is using keycloak for auth and identity management.
- Connect using different keycloak clients per service
- Use multiple keycloak instances if needed

🗣  Let's just speak about it but I think to avoid noise this proposal should not really include security part. 
We may have different proposal with options for that ( There are some proposals for keycloak already so we need to figure out how this will align with SDK)

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+}
+```
+
+#### Using Core Threads
+
+The Mobile Core SDK will provide Java Executors that wrap several common blocking tasks.  These executors will be available to any module or part of the application which may need to schedule diskIO work, netowrking work, or work to happen on the main thread.
+
+```
+AppExecutors.networkThread.execute(() -> {
```
wtrocki: If we going to use OkHttpd.enqueue method then is that still needed? 
Requires investigation on how we can handle requests/library
Worried about application lifecycle.

android-sdk/README.md
wtrocki: I have done example: https://gist.github.com/wtrocki/5774f883a9a2ae0b33c5aa63986cae45

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+  var result = someRequest.call(); //Perform a HTTP operation off the main thread
+  AppExecutors.mainThread.execute(()->{updateDisplay(result);});//Update the Android UI on the main thread.
+ });
+```
+
+#### Using Core Networking
+For now, networking will be a service exposed by a http service module.  The definition of this service will be a different proposal.
```
wtrocki: Spike ticket needed.

wtrocki:
Important things to add/consider from my point of view.
- Agreed over what should be the first service(outside proposal, but can influence it)
- What will be top level user usage for serviceX (for example sync) 
I had added 2 examples above, but this needs wider agreement. 
- Have separate chapter/proposal for security to reduce complexity and scope of this proposal.

wtrocki (dismissed):
I would like to get some of the elements cleared before we accept this.

wtrocki:
I'm going to pickup some of the spikes to streamline work on proposal.

android-sdk/README.md
secondsun: Yes it is still needed.  1) We don't want to expose OKHttp in our APIs and 2) we don't to encourage developers to add lots of threads of their own.

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Proposed solution
+
+### Core SDK
+
+The Core SDK will be the package that all service modules inherit from.  It will be responsible for managing network requests, threading, logging, error handling and resolution, loading configuration for mobile-core services, and other cross cutting concerns.  Modules will interface with the core SDK by implementing the necessary interfaces provided and annotating the implementations so that the Core SDK will be able to correctly manage them as well.  The Core SDK will also allow some of its internals to be exposed through interfaces for purposes such as testing, bespoke integrations, and management of additional resources.
+
+The Core SDK will provide a request lifecycle that the Psuedocode section will demonstrate.
```
josemigallas: Psuedocode -> Pseudocode

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Service Modules
+
+Service modules will each represent a singular service that is supportd by mobile-core.  We will use, where available and practical, existing libraries for services and in those cases the modules will provide the necessary implementation of Core APIs to make them compatable with the Core SDK lifecycles and features.  The first Services we are looking at integrations will be Keycloak, fh-sync-service, and a RESTful API service module.
+
+### Implementation notes
+
+Public APIs should always expose native Java Objects and not stringified JSON or JSON Objects.
+
+### Use cases and psuedo code examples
```
josemigallas: psuedo code -> pseudocode

android-sdk/README.md
josemigallas: So @secondsun an application module would be an Android Library module that I, as a developer, would have in my Android project, only locally?

android-sdk/README.md
josemigallas: What is this API service for? Like a proxy for an MCP API?

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+#### Starting up Mobile Core
+MobileCore is the "main" class for the SDK of the Core SDK.  Users instanciate it by placing a client configuration file in their assets directory, configuring MobileCore.Builder, and then calling its build method.
+
+```
```
josemigallas: Just an ornamental suggestion, adding the language (java) after the triple tick makes the snippets look much nicer.

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+Service modules will each represent a singular service that is supportd by mobile-core.  We will use, where available and practical, existing libraries for services and in those cases the modules will provide the necessary implementation of Core APIs to make them compatable with the Core SDK lifecycles and features.  The first Services we are looking at integrations will be Keycloak, fh-sync-service, and a RESTful API service module.
+
+### Implementation notes
+
+Public APIs should always expose native Java Objects and not stringified JSON or JSON Objects.
+
+### Use cases and psuedo code examples
+
+#### Starting up Mobile Core
+MobileCore is the "main" class for the SDK of the Core SDK.  Users instanciate it by placing a client configuration file in their assets directory, configuring MobileCore.Builder, and then calling its build method.
```
josemigallas: I guess it has to be called from the app's `Application` class? Would it be worth mentioning that here? Moreover, if the building is not called at all, would an exception be triggered or something?

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+Example KeyCloakService
+```
+public class KeyCloakService implements ServiceModule {
+   /* snip */
+   public KeyCloakService() {}
+
+   @Override
+    public void bootstrap(MobileCore core, ServiceConfiguration config) {
+        this.serverUrl = config.getProperty("auth-server-url");
```
josemigallas, line 94: I think there should be more consistency in property names, all of them either with dashes or low dashes. Also, to avoid typos the names could be declared as constants somewhere.

android-sdk/README.md
josemigallas: Is the 1st and 3rd parameters intended to be a harcoded String? I think the name should be declared as a constant or property somewhere. Maybe just a simple KeyCloakService.getType().

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+ServiceModuleRegistry.java
+```
+
+/**
+ * Registry of service modules.  It is responsible for maintaining a mapping of service modules to
+ * their types as well as the dependencies each module expects.
+ */
+public final class ServiceModuleRegistry {
+    private static final Map<String, Class<? extends ServiceModule>> serviceTypeMap = new HashMap<>();
+    private static final Map<String, String> dependencyMap = new HashMap<>();
```
josemigallas: Being dependencyMap a list of dependencies for the service, does it have to be a Map or could we leave it as just a List/Array?

android-sdk/README.md
josemigallas: Should the name be `dependencies` instead of `dependsOn`? It is clearer IMO as it implies it is a list of some kind.

android-sdk/README.md
josemigallas: Explicit is always better, but maybe we only need to define some Enum or List or group of constants that define the list of services registered in core. I don't know if it's good or bad approach to have all internal functionalities as "getters". For example what if I don't want to use Network or Keycloak in my application, then what happen if I invoke that functionality? (not a rhetorical question!). I guess the `getService` is more flexible from this point of view.
```java
core.getService(Services.NETWORK)
core.getService(Services.KEYCLOAK)
core.getService(Services.SYNC)
```

android-sdk/README.md
josemigallas: What approach exactly do you mean @wtrocki ?

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+#### Using Core Threads
+
+The Mobile Core SDK will provide Java Executors that wrap several common blocking tasks.  These executors will be available to any module or part of the application which may need to schedule diskIO work, netowrking work, or work to happen on the main thread.
+
+```
+AppExecutors.networkThread.execute(() -> {
+  var result = someRequest.call(); //Perform a HTTP operation off the main thread
+  AppExecutors.mainThread.execute(()->{updateDisplay(result);});//Update the Android UI on the main thread.
```
josemigallas: Doesn't Android have already a method to update the UI thread?

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+ });
+```
+
+#### Using Core Networking
```
josemigallas: Are we planning to implement out entire network library? Like our own Retrofit/Volley kind of thing?

android-sdk/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+    dependencyManagement {
+        imports {
+            mavenBom 'org.jboss.aerogear:aerogear-parent:1.0-SNAPSHOT'
+        }
+    }
+}
+```
+
+#### Android Studio Plugin
```
josemigallas: This would/will be very cool

android-sdk/README.md
secondsun: That is a debate for the future.

Ideally we will wrap our choice of implementation in a API so we can make behavior changes under the hood.

android-sdk/README.md
secondsun: Sort of.  You can statically get a reference to the mainLooper, but this is harder to test.  App Executors can follow a pattern like this : https://github.com/googlesamples/android-architecture/blob/todo-mvp/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/util/AppExecutors.java which lets us supply different executors so we can run tests outside of a emulator

android-sdk/README.md
josemigallas: OK that make more sense

android-sdk/README.md
secondsun: I have no strong feelings.

android-sdk/README.md
secondsun: No.  This code is actually wrong and this should be a multimap.
```
    private final Map<String, List<String>> dependencyMap = new HashMap<>();

```

android-sdk/README.md
secondsun: It is a hardcoded string for now, figuring out a typing mechanism is a problem for future us.

android-sdk/README.md
secondsun: I agree.  When we begin creating services we will have to revisit the issue.

android-sdk/README.md
secondsun: No not really.  It is up to each application's needs.  For instance someone may be building the MobileCore in a ApplicationScope dagger provider.

If building isn't called at all that means either someone is managing their modules manually or is going to have a NPE when they reference a module; either way an exception is not appropriate because they didn't use the builder.

android-sdk/README.md
secondsun: Because the SDK has a lifecycle for the modules to make sure the services are setup in order.  Your application module would be the code that runs last and depends on all of your services defined in mobile-services.json

android-sdk/README.md
secondsun: For service specific SDKs to be managed by the mobile core

android-sdk/README.md
secondsun: @josemigallas yes

android-sdk/README.md
secondsun: @josemigallas Basically.

david-martin (approved):
@wtrocki @wei-lee Could we get another approval (if all concerns are addressed or deferred) and I'll merge

android-sdk/README.md
josemigallas: Where the first String is the name of current service and the List its dependencies?

android-sdk/README.md
```diff
@@ -27,33 +31,60 @@ For example, if a KeyCloak and fh-sync-server service were created and bound, th
...
 
 Service modules will each represent a singular service that is supportd by mobile-core.  We will use, where available and practical, existing libraries for services and in those cases the modules will provide the necessary implementation of Core APIs to make them compatable with the Core SDK lifecycles and features.  The first Services we are looking at integrations will be Keycloak, fh-sync-service, and a RESTful API service module.
 
 ### Implementation notes
 
-Public APIs should always expose native Java Objects and not stringified JSON or JSON Objects.
+Public APIs should always expose native Java Objects and not stringified JSON or JSON Objects.  Our public APIs should not favor one library over another, instead we will abstract away our dependencies behind interfaces so that users may substitute their own and so we can change our implementations without breaking backwards compatability.
 
-### Use cases and psuedo code examples
+### Use cases and psuedocode examples
```
josemigallas: psuedo -> pseudo again 😁

android-sdk/README.md
wtrocki: Sync have reference to keycloak service.

android-sdk/README.md
secondsun: yes

david-martin:
@secondsun can you review feedback from remaining open comments.
We can merge if all comments have been addressed (either by explanation & agreement or by changes to the PR)

secondsun:
@david-martin I'll fix the typos that @josemigallas  picked out, I think that I have addressed @wtrocki 's comments, he and I can sync up on this.

wtrocki:
On last session  we agreed to not have dependencies between services themselves and provide top level example how SDK can be used in used application. We also agreed on supporting direct core api with injection, but proposal is missing that.  See my "-1 for this aproach." comment. 
I also agreed to close my PR as they changes should be included in original PR.

## Solution for the problem

I think it's worth to reduce scope of this and unblock PR's that may related on that. For example.
https://github.com/aerogear/aerogear-android-sdk/pull/12

Proposition is to have core module dealing with the configuration as library and make this configuration available for all services (by injecting core). 

This will mean to: 
- Remove service registry
- Remove dependency mechanism

After we get keycloak implementation and other services we could easily refactor common network elements into core. This will allow us to see what requirements and challenges core networking will require. Networking on android with OkHttpd is really simple and trying to write something sophisticated at the moment can be rabbit hole.

This suggestion will also allow us to deliver that functionality on time (within sprint range)

android-sdk/README.md
david-martin: @wtrocki I believe that is not the case. Looks like the name of the class was changed to `CustomService` to avoid confusion.

wtrocki:
After pair programming session with @secondsun we finalized elements around:

- Service registry
- Networking support

@secondsun is going to create PR's after that we can update proposal if needed and merge changes.

android-sdk/README.md
evanshortiss: Looks like I missed this stuff, but strongly agree with @wtrocki here. We don't want to place hard dependencies like this IMO

PR #10, opened 2018-01-10
Added proposal for PR testing guidelines

AdamSaleh:
This proposal should cover the PR testing of the repositories concerning the new Aerogear/Feedhenry developments.

It was based on a discussion between Adam Saleh, Vitali Chepeliuk, Pavel Sturc and Vojtech Sazel.

### DIFF
--- a/testing/pr-testing-guidelines-and-checklists.md
+++ b/testing/pr-testing-guidelines-and-checklists.md
@@ -1,0 +1,149 @@
+# PR testing guidelines and checklists 
+
+## Introduction
+
+This proposal should cover the PR testing of the repositories concerning the new Aerogear/Feedhenry developments.
+It was based on a discussion between Adam Saleh, Vitali Chepeliuk, Pavel Sturc and Vojtech Sazel.
+
+## Problem Description
+
+We currently don't have a proper set of guidelines for our PR checks and previously we used them mostly just to produce the build-artifacts
+in some fashion. The quality checks have often been added in ad-hoc manner and then disregarded/forgotten about.
+
+## Expectations
+
+If we create a set of coherent guidelines and model our automated PR checks by them we expect
+- to catch regressions early
+- to maintain a good quality of incoming code
+
+## Terms
+
+* relevant repository - repository that could influence ship-able artifacts by it's contents in a bad way. If we really don't want to follow these guidelines, just claim that the repository is currently irrelevant for our project.
+  
+## Proposed Solution
+
+These proposals only concern relevant repositories.
+
+### Repository should have a mandatory review by other person
+
+This is easily enabled in GitHub. It mandates that at least one person gives a PR a green check-mark.
+
+There will be some decisions that we don't want to automate on purpose, for example, we will probably display code-coverage prominently,
+but we won't have any automatically enforced that would stop PR from getting merged due to low coverage.
+
+In the future, we might have optional integration tests triggered by comments or specific tags on the PR. This would be hard to enforce automatically,
+so mandating those would fall as a responsibility of the reviewer.
+
+There could be a template with a checklist delivered by a .github/PULL_REQUEST_TEMPLATE.md
+Checklist would include:
+
+* Read through the code and ask about any unclear code
+* Try to checkout and run the relevant code
+* Evaluate if the new code is sufficiently tested, potentially consulting the code-coverage
+* Evaluate if there are any follow-up tasks to be logged
+* more?
+
+### All of the automated testing should finish in reasonable time, defined as in less than 10 minutes
+
+Looking at our previous PR checking attempts, even fast checks take around 2 minutes, due to inconsistencies in time of spawning our on-demand slaves.
+
+Our longest PR builds can take around 20 minutes.
+
+Any time, there is PR check that consistently hits the 10 minute run-time, it should become an issue in our backlog to be solved.
+
+### There should be a consistent and enforced style for our code
+
+Maintaining a repository where multiple developers collaborate should be helped by ensuring consistent style.
+While this often leads to a lot of bike-shedding, and changing the style after it has been applied can be tedious,
+but the style enforcement still seems to be quite useful, especially for new collaborators.
+
+For configuration files and dynamic languages, style checking can often find basic problems that would be discovered in compiled languages with type-checker.
+
+We could experiment even with automatic style formatting, such as:
+* gofmt
+* grunt-jsbeautifier
+
+We should enforce the style for:
+* go, probably with golint
+* javascript, probably with eslint, modeling after openshift-team's style guide
+* ansible, if some check exists
+* bash (proposed by Pavel Sturc)
+
+### If there is a build-able artifact, it should be built in reasonable subset of configurations
+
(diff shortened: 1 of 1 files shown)


testing/pr-testing-guidelines-and-checklists.md
```diff
@@ -0,0 +1,149 @@
...
+While this often leads to a lot of bike-shedding, and changing the style after it has been applied can be tedious,
+but the style enforcement still seems to be quite useful, especially for new collaborators.
+
+For configuration files and dynamic languages, style checking can often find basic problems that would be discovered in compiled languages with type-checker.
+
+We could experiment even with automatic style formatting, such as:
+* gofmt
+* grunt-jsbeautifier
+
+We should enforce the style for:
```
wojta: What would be the code styles? 
There is https://github.com/fheng/best_practice, but this is not public - We should move that to aerogear proposals as well.

testing/pr-testing-guidelines-and-checklists.md
```diff
@@ -0,0 +1,149 @@
...
+
+We probably want to include things like Docker-files here as well.
+
+### Unit tests should be run, results and coverage logged
+
+As we mentioned previously, coverage level shouldn't be enforced automatically, but reviewers should look at the coverage and test quality in general.
+
+### There should be at least some amount of integration testing, if at all possible
+
+ We should run our code in some sort of staging environment that would be reasonable compromise between development and production environment, with emphasis on speed of running of the test, i.e. we don't want to wait 20 minutes for openshift to spawn.
```
wojta: per PR this should be done only if it finishes in reasonable time

testing/pr-testing-guidelines-and-checklists.md
```diff
@@ -0,0 +1,149 @@
...
+
+* relevant repository - repository that could influence ship-able artifacts by it's contents in a bad way. If we really don't want to follow these guidelines, just claim that the repository is currently irrelevant for our project.
+  
+## Proposed Solution
+
+These proposals only concern relevant repositories.
+
+### Repository should have a mandatory review by other person
+
+This is easily enabled in GitHub. It mandates that at least one person gives a PR a green check-mark.
```
maleck13: As mentioned elsewhere I believe it was agreed that we would have 2 LGTM before merge, this stops rushed PRs and ensures knowledge of the change is spread across the team

testing/pr-testing-guidelines-and-checklists.md
```diff
@@ -0,0 +1,149 @@
...
+
+There will be some decisions that we don't want to automate on purpose, for example, we will probably display code-coverage prominently,
+but we won't have any automatically enforced that would stop PR from getting merged due to low coverage.
+
+In the future, we might have optional integration tests triggered by comments or specific tags on the PR. This would be hard to enforce automatically,
+so mandating those would fall as a responsibility of the reviewer.
+
+There could be a template with a checklist delivered by a .github/PULL_REQUEST_TEMPLATE.md
```
maleck13, line 39: PR template is a good idea 👍

testing/pr-testing-guidelines-and-checklists.md
```diff
@@ -0,0 +1,149 @@
...
+Checklist would include:
+
+* Read through the code and ask about any unclear code
+* Try to checkout and run the relevant code
```
maleck13: If we have good enough CI we shouldn't need this

testing/pr-testing-guidelines-and-checklists.md
```diff
@@ -0,0 +1,149 @@
...
+### There should be a consistent and enforced style for our code
+
+Maintaining a repository where multiple developers collaborate should be helped by ensuring consistent style.
+While this often leads to a lot of bike-shedding, and changing the style after it has been applied can be tedious,
+but the style enforcement still seems to be quite useful, especially for new collaborators.
+
+For configuration files and dynamic languages, style checking can often find basic problems that would be discovered in compiled languages with type-checker.
+
+We could experiment even with automatic style formatting, such as:
+* gofmt
```
maleck13: we will always be using gofmt, it is the only way to format go code.

maleck13: normally I use ```go vet``` ```go lint``` ```go fmt``` and ```errcheck``` plus tests before allowing PR to be merged (at least this is what most go projects do and what we did for mcp standalone)

testing/pr-testing-guidelines-and-checklists.md
```diff
@@ -0,0 +1,149 @@
...
+
+We should enforce the style for:
+* go, probably with golint
+* javascript, probably with eslint, modeling after openshift-team's style guide
+* ansible, if some check exists
+* bash (proposed by Pavel Sturc)
+
+### If there is a build-able artifact, it should be built in reasonable subset of configurations
+
+This mostly goes for binaries, such as our go-based cli tool, or compilable libraries for our mobile devices.
```
maleck13, line 80: by this do you mean build it for windows, mac and linux?

AdamSaleh: I am not sure. 

On one hand there is a great deal of utility even if we compile for just one target, i.e. type checker gets run.

Depends on our trust in the portability of the code. I think it might be fine to just build for linux (as that is what is easily accessible, are there cross compilation targets from linux to win/osx?) 

If we think there is use in having windows/osx builds on every PR (it is probable that most of our customers would use our tools from win/osx), then yes, we should do it.

maleck13: yes it is very easy to compile for different arch in go you should only need to set the env vars:
``` GOOS and GOARCH```

testing/pr-testing-guidelines-and-checklists.md
AdamSaleh: Sounds good. Unfortunately I didn't see a way to enforce two-people review on Github's protected branches. Do you think we should enforce it? There is a possibility to create another check that would use i.e https://developer.github.com/v3/pulls/reviews/ to check the reviewers, but I am not sure it is worth it.

Maybe I would drop the "enable in github" and just have it mentioned in the contributor guidelines and the PR template?

testing/pr-testing-guidelines-and-checklists.md
AdamSaleh: True, will reword this to specifically mandate we have good enough CI that reviewers don't need to run the code :-)

testing/pr-testing-guidelines-and-checklists.md
AdamSaleh: Awesome, in that case we should definitely do that.

Do you think we should run the tests on multiple arch as well? That might be the tricky part (I don't really want to maintain windows slaves unless necesary )

maleck13: No I see no need for that

maleck13:
@AdamSaleh I think once you have updated this with the changes from the call it is good to merge as there was consensus  on the call.

wojta:
Was this updated?

PR #13, opened 2018-01-18
proposal for shared services

philbrookes:


apbs/use-shared-services-across-namespaces.md
```diff
@@ -0,0 +1,54 @@
+# Abstract
+
+APBs for resource intensive services should be programmed to re-use the service that they would otherwise deploy, if that service is already deployed and available in another namespace.
```
maleck13: might want to add a little more to the definition (the service need to support being shared) For instance at the moment I don't think there is anyway to share the sync service

apbs/use-shared-services-across-namespaces.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Terms
+
+- APB ([ansible playbook bundle](https://docs.openshift.org/latest/apb_devel/index.html))
+- Shared Service: A service that is used across multiple namespaces
```
maleck13: A service that can be used across multiple namespaces

apbs/use-shared-services-across-namespaces.md
```diff
@@ -0,0 +1,54 @@
...
+- System Resources: CPU or memory
+- Namespace / Project: These refer to the same thing and may be used interchangably.
+- apb.yml: ([APB spec file](https://docs.openshift.org/latest/apb_devel/writing/reference.html#apb-devel-writing-ref-spec)) Configures the inputs passed to an APB prior to deployment.
+
+## Problem Description
+
+Some of the services used in the development of a mobile application can require quite a lot of system resources which could turn into a factor that limits adoption of the Mobile Platform.
+
+For example, if multiple developers are building a product which requires KeyCloak, and they do not have the resources available to run a seperate KeyCloak instance in each developers project, there is currently no convenient way for all of these developers to develop their application using a single instance of KeyCloak.
```
maleck13: Change Keycloak to Authentication Service

apbs/use-shared-services-across-namespaces.md
```diff
@@ -0,0 +1,54 @@
...
+- The APB for a shared service automatically configuring the shared service in another namespace and exposing the configuration details to the current namespace.
+
+## Non-Goals
+
+- The platform intelligently discovering and connecting to services that are running in seperate namespaces.
+- The platform intelligently discovering and suggesting services to connect to.
+
+## Proposed Solution
+
+For services that are suited to being shared services (such as KeyCloak), we can make a change the apb.yml, allowing the developer to specify connection details for an already existing instance of the service this APB would otherwise deploy.
```
maleck13: Should we specify that these are specifically provision parameters and maybe given an example of what they look like

apbs/use-shared-services-across-namespaces.md
```diff
@@ -0,0 +1,54 @@
...
+
+Example:
+![Image of KeyCloak APB config screen with shared service fields](./images/shared-services-config.png)
+
+In this example, if the `Host of existing KeyCloak service` were left empty then a new service would be installed. However if this field were populated the APB would connect to the host with the provided auth details and create a configuration as required.
```
maleck13: > and create a configuration as required.

And run through the rest of the APB as though the instance had finished provisioning the actual running instance

maleck13:
@philbrookes I think this is a good start, however I think we should also cover how a binding will work for a shared service. For instance, if you had a shared keycloak and then performed a binding against that serviceinstance that is pointing to a service in a different namespace. I don't think there will be too many difficulties but I think we should cover it in the proposal

wei-lee (commented):
@philbrookes @maleck13 I have a few questions. If this is enabled, is it possible to have this situation:

* In namespace A, fh-sync service is bound to KC in namepsace B
* In namespace A, metrics service (or maybe another service) is bound to KC in namespace C

So we end up with 2 services are bound to completely different KC instances.

First of all, should we allow this to happen?

Secondly, if this is the case, how should we deal with it in the auth SDK? At the moment, the auth SDK only works across services that are protected by the same KC that the user is authenticated with.

maleck13:
> In namespace A, metrics service (or maybe another service) is bound to KC in namespace C

@wei-lee  I don't think this will happen.
When provisioning the second KC it there will be a new serivceInstance created in namespace A that refers to the KC in namespace B. When asking for a binding within a namespace, it uses the service instance in that namespace as a reference. So  it will be a binding to the keycloak that the serviceintance is associated with. In this case the one in namespace B.

@philbrookes  I think we have to consider what may happen though if a developer deletes the serviceinstance in Namespace A, the deprovision playbook needs to be intelligent enough to only remove the realm and secrets / configmaps and not the the actual running KC instance

apbs/use-shared-services-across-namespaces.md
philbrookes: I've attached an image below this paragraph showing what they might look like, is that what you meant?

wei-lee:
@maleck13 ah, right. Gotcha. There will still be a KC instance in the namespace A which should be used for binding by other services. But this instance maybe a pointer to a instance in another namespace. Is it right to say?

maleck13:
@wei-lee there will be a serviceinstance in namespace A yes that serviceinstance is "backed" by the KC in  namespace B. Any bindings that happen in namespace A will be directed to KC in namespace B

maleck13:
Also needs to cover in detail how a deprovision on a shared service should behave. (spoke with Phil about this on a call). Proposal should cover how each action that can be performed against a regular instance will now work against a shared service

apbs/use-shared-services-across-namespaces.md
```diff
@@ -0,0 +1,54 @@
+# Abstract
+
+When a service is suitable for sharing across namespaces and already deployed and available in another namespace, the APB for deploying those services should be programmed to re-use the service that is available rather than deploying another instance of it.
```
david-martin: Are we limiting users to services already deployed to a namespace? What about services running outside of kubernetes?

apbs/use-shared-services-across-namespaces.md
```diff
@@ -0,0 +1,54 @@
...
+
+- APB ([ansible playbook bundle](https://docs.openshift.org/latest/apb_devel/index.html))
+- Shared Service: A service that may be across multiple namespaces
+- System Resources: CPU or memory
+- Namespace / Project: These refer to the same thing and may be used interchangably.
+- apb.yml: ([APB spec file](https://docs.openshift.org/latest/apb_devel/writing/reference.html#apb-devel-writing-ref-spec)) Configures the inputs passed to an APB prior to deployment.
+
+## Problem Description
+
+Some of the services used in the development of a mobile application can require quite a lot of system resources which could turn into a factor that limits adoption of the Mobile Platform.
```
david-martin: suggesting "adoption of Mobile Services" instead of "adoption of the Mobile Platform"

apbs/use-shared-services-across-namespaces.md
david-martin: Image gives a good example :+1:

apbs/use-shared-services-across-namespaces.md
```diff
@@ -0,0 +1,54 @@
...
+For services that are suited to being shared services (such as KeyCloak), we can make a change the apb.yml, allowing the developer to specify connection details for an already existing instance of the service this APB would otherwise deploy.
+
+Example:
+![Image of KeyCloak APB config screen with shared service fields](./images/shared-services-config.png)
+
+In this example, if the `Host of existing KeyCloak service` were left empty then a new service would be installed. However if this field were populated the APB would connect to the host with the provided auth details and then complete the APB as usual using this existing shared Service.
+
+Once the shared service has been configured via the APB, the APB will then expose these configuration details to the current namespace in it's usual manner (i.e. config-map, secret, etc).
+
+Should the APB be unable to connect to the supplied Shared Service details then the APB should exit with a failure status and report the problem in a way that will make trouble-shooting as convenient as possible for the developer.
```
david-martin: Should/If

apbs/use-shared-services-across-namespaces.md
```diff
@@ -0,0 +1,54 @@
...
+- UPS
+- Build Farm
+
+### Potential Difficulties
+
+#### Cluttered Configuration Screen
+The form for configuring the shared service via the APB could start looking quite busy with the extra inputs, an ideal solution to this would be fields that display conditionally with a check-box like "This is a shared service" which would hide the deploy options and instead display the shared service options, at the moment this is not possible, but having contacted the ASB team they have informed that it is on their roadmap.
+
+#### Unclear Failures
+Currently, when an APB fails to deploy, the error can be quite hard to find (you need to look at the APB pod logs in the APB specific namespace - which might not be visible to the currently logged in user)
```
david-martin: Is there upstream work/proposals/roadmap items that would address this?
Good to call this out, but it's not specific to the shared services scenario. Its a potential difficulty with any APB.

david-martin (changes requested):
@philbrookes few comments. Nothing major

apbs/use-shared-services-across-namespaces.md
```diff
@@ -0,0 +1,54 @@
...
+
+Some of the services used in the development of a mobile application can require quite a lot of system resources which could turn into a factor that limits adoption of the Mobile Platform.
+
+For example, if multiple developers are building a product which requires an Authentication Service, and they do not have the resources available to run a seperate instance of an Authentication Service in each developers project, there is currently no convenient way for all of these developers to develop their application using a single shared Authentication Service.
+
+## Goals
+
+- A convenient way for the developer to tell the platform about an existing deployment of the service he is about to deploy.
+- The APB for a shared service automatically connecting to the shared service in another namespace, using the provided details.
+- The APB for a shared service automatically configuring the shared service in another namespace and exposing the configuration details to the current namespace.
```
maleck13: Do we want to add deprovision and bind goals here?

apbs/use-shared-services-across-namespaces.md
```diff
@@ -0,0 +1,70 @@
...
+- KeyCloak
+- UPS
+- Build Farm
+
+### Deprovision
+When running a deprovision of Connected Service, the APB should clear up it's own configuration from the shared service, but should not deprovision the shared service.
+
+For example, in namespace-a the KeyCloak APB would have connected to KeyCloak in namespace-b and created a realm, a client, and a bearer-client. This realm and users should be removed from KeyCloak in namespace-b and the secret they are stored in, in namespace-a. Keycloak in namespace-b will still be running when this is completed.
+
+When deprovisioning a Shared Service the service should be removed whether or not there are still existing Connected Services.
```
maleck13: I think we need to make this a little bit clearer, I know what you mean but could be hard to parse.
So essentially if the resources created during the provision are present in the namespace then it will deprovision them otherwise it will only remove service specific items (such as the clients etc) and not the actual running pods

maleck13 (changes requested):
2 small comments but looking good now

maleck13 (approved):

maleck13:
@david-martin you ok to give it another look over and approve if you are happy

david-martin (approved):

david-martin:
I'm happy with the changes after feedback :+1:  Merging

PR #14, opened 2018-01-23
Mobile App Metrics and SDK version tracking proposal

paolobueno:
## Description
Gathering information from the related spikes and docs.

Branch created in upstream repository instead of own repository to allow for contributions from others.

## Tasks

- [x] service architecture
- [x] how the service works
- [x] how the SDK works
- [x] expected initial API (AEROGEAR-1894)
- [x] Storage solution evaluations (AEROGEAR-1894)
- [x] how data can be queried and visualised
- [x] extra modifications to grafana from the current metrics-apb (AEROGEAR-1892)
- [X] investigate mongo
- [x] investigate influx


metrics/glossary.md
```diff
@@ -0,0 +1,11 @@
+## Glossary
+
+- Metrics - Arbitrary measurements that represents current or historical data from a Service, App or User. Some examples:
```
wtrocki: Metrics - are usually connected to some changing values: like performance metrics etc.
In this case statistics may be better name.

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,131 @@
+# Abstract
+
+This document contains the proposal for gathering metrics on system events for running Mobile Apps and making them available to developers.
```
wtrocki: [Optional] [Suggestion] In this case statistics may be better name.

wtrocki:
@paolobueno Reviewed. Good job. We can update that once we will settle on what initial storage solution will be used.

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,131 @@
...
+
+##### Pros
+
+- Shared service with backend metrics
+- Requires an additional service for long-term metrics storage
+
+##### Cons
+
+- Data model is time-series exclusive
+- Pull-based data gathering
```
aliok: Although the convenient and common way of data gathering in Prometheus is pull-based, pushing data to a gateway is also possible. https://prometheus.io/docs/instrumenting/pushing/
I would hate to see that gateway in use though :smile:

aliok:
@paolobueno great work!

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,131 @@
...
+
+- Registering of user-defined metrics metrics
+
+## Terms
+
+For a list of terms used in this document, see [Glossary](./glossary.md).
+
+## Problem Description
+
+As mobile applications become more successful, system metrics data becomes invaluable for developers to observe the landscape of application installations and make sure further development and deployment of backend services will not affect users.
```
david-martin: This is good.
A small change I'd recommend
"As Mobile Apps are developed, improved and iterated on, different versions of those Apps can be used on many different devices. System metrics data ..."

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,131 @@
...
+
+- Development and maintenance of the new service;
+
+- Types of gathered metrics can become a subset of the underlying Storage's full range of types.
+
+<!-- TODO: mention logstash, fluentd -->
+
+### Storage
+
+The Storage service will store metrics data forwarded from the Mobile Metrics Service. Its implementation can be supplied by the following projects:
```
david-martin: I'd like to see this section be much smaller, focusing on the solution we're proposing based on any investigations etc..
It could link off to any jiras where the investigation was done, and conclusions were made.

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,131 @@
...
+- No clustering support in OSS offering
+- Medium-sized minimum resource requirements
+
+### Querying and Data Visualisation
+
+The Querying and Visualization service will be supplied by Grafana, which can utilize the Storage service implementation directly as a Data Source.
+
+User interface familiarity is an important factor for developers that use the overall metrics service, so addition of other visualisation tools like Kibana that would increase the learning curve of the solution should be avoided.
+
+#### Authentication
```
david-martin: @paolobueno I think we should have something about api key auth for mobile apps calling the "thin HTTP-based service". I don't think we want those endpoints to be public.

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,131 @@
...
+
+##### Pros
+
+- Long-term storage adapter for Prometheus
+- HTTP-based API for push-based metrics gathering
+- Flexible querying with direct comparison to SQL
+
+##### Cons
+
+- Focused on time-series data
```
aliok: @paolobueno do you have an idea if text search is possible? I am thinking about remote logging feature we are going to provide somewhere in the future.

aliok: ah, ok, saw it here: https://docs.influxdata.com/influxdb/v0.8/api/query_language/
`select * from events where email !~ /.*gmail.*/;`

aliok: hmm, logs are perhaps not natural to store in InfluxDB though. we might need some additional research if Influx can be as data storage for remote logging. this is low priority for now.

david-martin: agreed on priority. Remote logging can be a separate proposal (but may or may not re-use some elements in this proposal).
There are no high level stories/trellos for a Remote Logging service at this time (but good to be aware of it long term)

metrics/mobile-app-metrics.md
paolobueno: Agreed!

@aliok I'm going to move the contents of this section + link the google doc to the [spike issue](https://issues.jboss.org/browse/AEROGEAR-1894) then.

metrics/mobile-app-metrics.md
paolobueno: Sure, this one is a subheader for the "Querying and Data Visualisation" section, so I'll make it clear this relates to only acessing the grafana web ui and add another auth subheader to the "Mobile Metrics Service" section.

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,108 @@
+# Abstract
+
+This document contains the proposal for gathering metrics on system statistics for running Mobile Apps and making them available to developers.
+
+## Goals
+
+### Primary Goals
+
+- Provide mobile app developers a service that can receive overall runtime metrics (processor and memory data, authentication events, runtime versions, networking events, etc).
```
aliok: @david-martin @wtrocki @pb82 
Should we generalize things? Or just talk about mobile app version and sdk version metrics?

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+
+- Provide developers an interface to query and visualize the gathered data in order to make informed decisions towards deployment, deprecation of backend services and other actions that depend on the state of an application's installed base.
+
+### Secondary Goals
+
+- Provide mobile app developers with an easy to use SDK to connect easily and securely to the metrics service
```
aliok: @david-martin @wtrocki @pb82
Should we think about security?

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+
+The following diagram illustrates the components of the Metrics Service:
+
+![service diagram](./app-service-diagram.jpg)
+
+### Mobile App Client
+
+A native, platform-specific SDK for targetting the Metrics Service will be available to developers.
+
+Upon installation into a project and providing configuration for targetting the Metrics Service the SDK will automatically send general information about the general runtime for this application with a unique device ID, namely the `[SDK Version, clientID, timestamp]` tuple upon application initialization.
```
aliok: @david-martin @wtrocki @pb82
Upon installation or after every app start? I think to answer "how many app starts happened in last N days with sdk version 1.2.3?", we need to send it every app start.

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+
+The SDK will also instrument calls at the network level in order to gather metrics around communication and bandwith consumption.
```
aliok: @david-martin @wtrocki @pb82
should we have this?

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+
+Other system resources utilization data can also be sent periodically.
+
+### Mobile Metrics Service
+
+A thin HTTP-based service will be present to receive metrics data from the mobile applications in an storage-agnostic way, then forwarding to the Storage service.
```
aliok: @david-martin @wtrocki @pb82
Basically, we can have different implementations for different storage options. But I can't really call that storage agnostic.

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+
+Compared to mobile applications interfacing directly with the Storage service, this presents the following tradeoff:
+
+#### Pros
+
+- Possibility to upgrade or change the Storage service independently of mobile client SDK version and implementation;
+
+- Possible support for different Storage engines which can be selected based on application needs;
```
aliok: @david-martin @wtrocki @pb82
What about this idea?

metrics/mobile-app-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+
+`POST /sdk-version`
+
+Receives `Content-Type: application/json` containing the SDK Version and a clientID that should be called from the SDK initialization process. A timestamp can be generated server-side.
+
+### Authentication for the Mobile Metrics Service
+
+The HTTP endpoints for the Mobile Metrics Service is going to be accessbile only via HTTPS to devices that provide an authentication header containing an appropriate API Key.
+
+This key will be acessible to the SDK via new fields in the generated `mobile-services.json` from the mobile CLI.
```
aliok: @david-martin @wtrocki @pb82
Should we think about this right now? Is it part of the epic?

aliok:
@david-martin @wtrocki @pb82 
I asked a couple of questions inline. I was reading the great proposal @paolobueno wrote and wanted to clear some stuff before updating the document with the new architecture.

metrics/mobile-app-metrics.md
pb82: Maybe you can leave the generalization but mention that initially we will only record SDK and APP version?

metrics/mobile-app-metrics.md
pb82: At some point we will have to gather some ideas about securing the metrics endpoint. I assume it can't be linked to app auth because we want to send the initial metrics data as soon as the app launches. So it's probably something like https + API key?

I think that for now this can be left unspecified but it might be something we need to think about in the next sprint.

metrics/mobile-app-metrics.md
pb82: Yes we need app start to get data about current usage. Not sure if it even makes sense to send a metrics event on install: it would make sense if we also recorded uninstall (to get the accurate numbers of installations). But as far as i know it's not possible for an App to know when it's uninstalled. So maybe something like `distinct(clientId)` for the total number of installations?

metrics/mobile-app-metrics.md
pb82: What about: `A thin HTTP-based service will be present to receive metrics data from the mobile applications and abstract the storage layer, ...` ?

metrics/mobile-app-metrics.md
pb82: There is no card on the trello board about securing the sdk metrics. @david-martin should we add a card for this?

metrics/mobile-app-metrics.md
pb82: Our Trello board currently does not mention this. Only version distribution. But probably something for the backlog.

metrics/mobile-app-metrics.md
pb82: Don't think this should be in the proposal. Maybe this was written when we were thinking about Cluster level/Namespace level elastic instance?

aliok:
@david-martin @pb82 @wtrocki 
Note that client side of things will not be included in this proposal.
Client side seems in a separate epic: 
- https://trello.com/c/lt4jc3h4/10-as-a-developer-i-want-to-be-able-to-visualise-runtime-metrics-about-what-versions-of-the-mobile-service-sdks-are-in-use-so-that
- https://trello.com/c/BZ4Y8WBB/24-as-an-android-developer-i-want-a-metrics-sdk-as-part-of-the-overall-android-sdk-so-that-the-sdk-will-capture-metrics-information

metrics/mobile-app-metrics.md
wtrocki: +100 for @pb82 suggestion - IMHO proposal should be clear and explain clearly what technologies will be used and implications etc. so people will be able to use it for implementation without "leaving any doors open".

metrics/mobile-app-metrics.md
wtrocki: @pb82 That's really good topic. Let's kickstart discussion on aerogear list as this may be out of the scope of this document and may be lost, with edits.

metrics/mobile-app-metrics.md
wtrocki: `mobile-services.json` file could contain token for each individual app and  this token can be supplied for every metrics request. Token will allow to push metrics only. 

Sending email to list to gather more opinions

aliok:
cc @david-martin @wtrocki @pb82 @darahayes @StevenTobin 
Updated the proposal. Didn't want to go into too much detail in complex technical things as they're captured in spike JIRAs.

metrics/default-mobile-app-metrics.md
```diff
@@ -0,0 +1,159 @@
...
+
+Please note that, the client side is not part of this proposal.
+
+### Mobile Metrics Service
+
+A thin HTTP-based service will be present to receive metrics data from the mobile applications and then to forward to the storage service.
+
+At a PoC level, the endpoints available from the Mobile Metrics Service should be:
+
+`POST /default-metrics`
```
darahayes, line 55: Don't know how applicable this is but I have a suggestion. Would it make sense to version this API? e.g. something like `/api/1.0/default-metrics`?

I feel like the metrics will change a lot over time. Just a suggestion.

darahayes (approved):
Just had one comment which may not be relevant at all. Other than that it looking good!

metrics/default-mobile-app-metrics.md
david-martin: @darahayes Good point. 
However, I think for and MVP of a metrics service this may be adding unnecessary complexity to the service and SDKs.

metrics/default-mobile-app-metrics.md
```diff
@@ -0,0 +1,159 @@
...
+}
+```
+
+It is not part of this proposal what client id is, but it must be something unique per device/client.
+
+### Storage
+
+Prometheus will be used for time series data. However, for long term storage and better scaling, Prometheus will use Postgres as the target long term storage.
+
+For non-time series data, metrics service will directly talk to Postgres.
```
david-martin, line 73: @aliok Was it not the case that the metris service would always talk directly to Postgres?
That would eliminate the need for a prometheus 'pull' service or similar

david-martin: Looking at the architecture diagram, I see prometheus would be pulling from the metrics service (via an auto discovered metrics endpoint?)
Does that mean there is no intermediate process between the metrics service and prometheus?
Does that mean the metrics service needs to temporarily store metrics (in memory) untill prometheus polls?

david-martin:
@aliok I think we're close to merging.
I've left a couple of comments if you could review

metrics/default-mobile-app-metrics.md
pb82: @david-martin I don't think that's a problem. The metrics service wouldn't really store any records in memory. If we use the Prometheus client libs (for Golang, Java or Node) then it's scraping endpoint only exposes a number of counters and gauges. The records (or rather the timeseries) only exist on the Prometheus side and are appended after each scraping.

pb82: @aliok @david-martin On the other hand there is not really a reason for the metrics service to be scraped by Prometheus: this service does not deal with time serices data, it only allows SDK metrics to be pushed. So there is probably no need to expose the data to Prometheus and the service should write to Postgres directly.

metrics/default-mobile-app-metrics.md
aliok: @david-martin I can see an "API versioning" card in product Trello board (https://trello.com/c/71UiKc1F/12-as-a-developer-i-want-to-be-able-to-version-my-apis-and-manage-different-versions-for-different-sets-of-client-apps-i-will-need). But it is about versioning custom user APIs. I don't see anything about versioning SDK API versioning. You might want to talk about this with @johnfriz for future.

@darahayes nice catch!

metrics/default-mobile-app-metrics.md
aliok: @david-martin @pb82 
Prometheus would be pulling the time series data (sdk init events stream, not <client,sdkVersion> map) from the service. There is no process between metrics service and Prometheus.
As @pb82 said, like in backend metrics, service will increase counters (with labels etc) provided by Prometheus client library which basically means storing them in memory.

Our target was to use Prometheus for time series data. We can of course send those directly to Postgres, but in that case we won't use Prometheus. Theoretically, things could be inserted in a way that's compatible with Prometheus, but I don't see any point there.

aliok:
@david-martin I think we're in a state that we can share the proposal with the wider group. Doing it now.

metrics/default-mobile-app-metrics.md
darahayes: @david-martin @aliok @pb82  I'll start by making some distinctions so nobody misunderstands me. For the sake of this conversation there are two types of metrics:

* Business metrics - The metrics we collect from the mobile clients. SDK version, OS versions etc.
* System metrics - metrics about the collector service itself. Basic things like memory usage, CPU etc. The stuff that's normally provided by whatever prometheus library we include in the service. (as @pb82 mentioned)

I was under the impression that the collector service was going to place all business metrics directly in postgres (timeseries & non-timeseries).

I figured there would still be a metrics endpoint on the collector service but but just for the collector service's system metrics. 

Not sure what benefit there is having the collector service expose our business metrics to prometheus via an endpoint **and** via postgres. It adds an extra layer of complexity because we will need to store the business metrics in memory (like @david-martin  mentioned). There's a much greater risk of data loss and inaccurate metrics with this approach. 

I think we're forcing ourselves into a corner with this idea that `time-series == prometheus` and `non time-series === postgres`

Are there any advantages to this approach. Am I missing something here? I'd love to get your ideas.

pb82: @darahayes @aliok @david-martin Isn't the whole point of Postgres allowing us to store Business metrics (client sdk metrics)? Those metrics can't be exposed to Prometheus and Prometheus can't store them in a way that is useful to us. Example: which clientId uses which Sdk version.

So i also don't think there is a point in having a Prometheus endpoint on the service (which isn't really a collector service either: we exclusively push to it). Collecting metrics from system services (Keycloak etc.) on the other hand is still the job of Prometheus' scraper.

aliok: @darahayes @pb82 

Business metrics is 2 parts:
1. Time series ones: event streams
2. Non-time series ones: maintained map

(1) goes to Postgres via Prometheus. (2) goes to Postgres directly.

> we will need to store the business metrics in memory (like @david-martin mentioned). There's a much greater risk of data loss and inaccurate metrics with this approach.

I wouldn't worry too much on this. Keeping them in memory is fine as scraping happens often.

> (which isn't really a collector service either: we exclusively push to it)

We actually don't push to Prometheus. We increase gauges in the service application and Prometheus scrapes them.

The initial idea was to use Prometheus for business metrics. And we would anyway need something like Postgres for long term storage as Prometheus sucks when it comes to that.
Then we figured out that we need non-time series data that we need to maintain (<clientId,sdkVersion> map). As Prometheus doesn't support that, we thought why don't we use the same Postgres that's used for long term storage of time series data, for non-time series data as well.

We could actually eliminate Prometheus from the picture. Then we need to push time series data to Posgtres, like non-time series data. This would need a POC though.
This is architecture#1 we discussed in this JIRA ticket: https://issues.jboss.org/browse/AEROGEAR-1894

We thought architecture#4 in that ticket, the one that's described in this proposal, is a better approach. @david-martin maybe you can tell us more why this one is preferred?

I am not a big advocate on any architecture. I am OK with both :)

darahayes: I'm happy with the current architecture and prometheus should definitely be kept. We're not just using prometheus for business metrics. It's being used to scrape important low level system metrics from all the other services running in openshift. It might even be used to expose OpenShift level metrics in the future also.

My question is about the business metrics. You say:

> (1) goes to Postgres via Prometheus. (2) goes to Postgres directly.

My question is why? I'm failing to see the advantage of this. Why can't we store both directly in postgres?

david-martin: I think storing both directly in Postgres from the App Metrics service would make more sense architecturally.
The App Metrics service would *not* be tightly coupled with Prometheus then.
 
There is still benefit in having both the App Metrics bits and the Service Metrics bits all in the 1 APB.
The Service Metrics can use Postgres (from the App Metrics) for long term storage.
The App Metrics can use Grafana (from the Service Metrics) for visualisation.

darahayes: I think that makes the most sense. Although it's all being bundled together in the APB, there is a clear separation of concerns. For example it would technically be possible to deploy a fully functioning App metrics service + postgres and not be reliant on prometheus. (Not that I'm suggesting this as an option).

pb82: @david-martin I don't know if that can be done easily. We also use Postgres as the storage provider for Prometheus (with Timescale Db?). So if you want the metrics service to put timeseries data directly into Postgres you would have to conform to whatever format Timescale Db expects. That might be easy, i don't know. 

Having the metrics service expose time series data through a prometheus endpoint is not an issue in my Opinion: I could be wrong but i don't think it stores anything in memory other than the immediate values of the counters and gauges. Pros: we would have one interface for all time series data.

The metrics service still has to put non-timeseries data directly into Postgres. So it's not tightly coupled: if you don't have Prometheus you still get the non-timeseries metrics.

aliok: > So if you want the metrics service to put timeseries data directly into Postgres you would have to conform to whatever format Timescale Db expects. That might be easy, i don't know.

this is not a problem.

> Pros: we would have one interface for all time series data.

+1. Yes this is the actual pro.

aliok:
Closing this one. Gonna create a new proposal for architecture#1 here: https://issues.jboss.org/browse/AEROGEAR-1894

PR #15, opened 2018-01-24
Add evaluation notes for MongoDB

aliok:
ping @paolobueno

### DIFF
--- a/metrics/mobile-app-metrics.md
+++ b/metrics/mobile-app-metrics.md
@@ -91,7 +91,7 @@ The Storage service will store metrics data forwarded from the Mobile Metrics Se
 ##### Pros
 
 - High advertised performance
-- Highest variety of data types and modelling
+- Highest variety of data types and modelling (supports time series data and non-time series data)
 - Rich querying possibilities
 
 ##### Cons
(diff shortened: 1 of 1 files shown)


metrics/mobile-app-metrics.md
```diff
@@ -100,12 +100,16 @@ The Storage service will store metrics data forwarded from the Mobile Metrics Se
...
-<!-- TODO -->
-
 ##### Pros
 
+- Supports time series data and non-time series data
+
 ##### Cons
 
+- Support for time series data is not native thus needs modeling work
+- Grafana/Kibana cannot talk to MongoDB without ugly workarounds
```
wtrocki, line 110: Good points. Definitely using Mongo with Grafana/Prometheus is no go zone.  
I got impression that when using mongodb we could develop simple panel as part of the service to showcase charts etc.

aliok: @wtrocki +1, it is a good alternative.
But, if we go that way and have some application with business logic, I would prefer implementing an endpoint that Grafana can talk to. Let's first evaluate the current idea and then we can discuss alternatives like that.

paolobueno (approved):

PR #16, opened 2018-01-24
Mobile configuration standards

wtrocki:
Really important topic for the SDK maintenance and versioning. 
I do not have strong opinions about any of the suggested formats - it's just more important to settle on something that will be structured and make sense from mobile point of view.

### Motivation

Hide service specific configuration in metadata. 
Expose common elements to top level to not mix them with metadata.


philipgough:
@wtrocki I don't have enough oversight to have any input on what the service object should look like but not including internals and standardising on a specific format would seem like a no brainer so +1

Mobile-configuration.md
```diff
@@ -0,0 +1,62 @@
...
+Introduce standard for system interoperability, by providing schemas that will be used to communicate between mobile sdk's and core services.
+
+## Problem Description
+
+Mobile configuration format is not standarized. It contains various fields that have no meaning for the mobile users.
+Configuration also exposes some internal fields. 
+
+```json
+{
+ "config": {
```
maleck13: I believe this issue is addressed in a previous proposal  https://github.com/aerogear/proposals/blob/master/apbs/create-secret-and-configmap-during-provision.md
Public information is stored (will be stored) in a configmap private information in a secret (the configmap is what will be turned into the configuration for the client).

wtrocki: There are multiple issues here. Just pointed out the most prominent ones. 
I could contribute changes to CLI just to sort out this stuff as it causing headaches when trying to implement SDK's.

maleck13:
Thanks for starting this @wtrocki We have a task in the next sprint to do something similar see https://trello.com/c/SNNkv5pH/41-5-as-a-cli-developer-i-want-to-be-able-to-retrieve-my-sdk-config-in-json-format-so-that-i-can-include-it-in-my-mobile-applicatio

If you are ok with it we can use this as a base to complete part of the requirement to have the format well defined and documented, it will also be documented in mobile-core

Mobile-configuration.md
```diff
@@ -0,0 +1,65 @@
...
+## Proposed Solution
+
+Provide standard for mobile configuration that will be extendable and have backwards compatibility.
+
+### Mobile cli specifics
+
+Mobile-cli currently print configuration out to the std. 
+Mobile-cli should have option to create `mobile-services.json` file in specified location.
+
+### Json format specific
```
wtrocki: @maleck13 - Is that making a sense?
Motivation is to get service specific elements to one common object that can be Dictionary or HashMap. 
Add version (really important in my opinion)
Split between name, type and id - they are totally different fields. 
This will allow people to provision more than one service of same type (if that will be needed) and this will not crash SDK.

wtrocki:
@maleck13 It will be good to keep SDK team in the loop so we could point how this may work from mobile perspective. Thanks for taking look

Mobile-configuration.md
maleck13: well really what we need is for the keycloak apb to implement the changes outlined in the proposal. IE put the client config stuff into a config map and other stuff into the secret. 
I am happy to have changes contributed to the CLI, just be aware we are going to be doing a lot of work around this area next sprint

wtrocki: @maleck13 - I think that we should hold off with implementation till you get all requirements. For example: For the moment keycloak is represented as service, but there may be use cases where individual service configurations will need to have some keycloak specific data etc.

Implementation for this format is trivial, but I would like to just have this PR and stage changes we are trying to implement so both SDK and Core team will have common place to innovate. Once we are happy with outcome (at some point of sprint) we could implement the changes.

Mobile-configuration.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+### Service object
+
+{
+  "id": "string",
+  "name": "string",
+  "type": "string",
+  "url": "string",
+  "metadata": "array"
```
paolobueno: This sounds like it should be `object`. So for the mysql example above it'd be 

```json
{
...
"metadata": {
  "headers": {},
  "mysql_db": "unifiedpush",
  "mysql_password": "unifiedpush",
  "mysql_username": "unifiedpush"
}
}
```
right?

Mobile-configuration.md
```diff
@@ -0,0 +1,66 @@
...
+### Service object
+
+```json
+{
+  "id": "string",
+  "name": "string",
+  "type": "string",
+  "url": "string",
+  "metadata": "object"
+}
```
secondsun: You know, it might be worth making a schema document that we can export to tons of different languages for tooling purposes.

Mobile-configuration.md
```diff
@@ -0,0 +1,67 @@
...
+          "mysql_username": "unifiedpush",
+          "name": "unified-push-server",
+          "type": "unified-push-server",
+          "uri": "http://ups-myproject.192.168.37.1.nip.io"
+        },
+        "name": "unified-push-server"
+}
+```
+
+Most important parameters often have different names. For example some services contain `url` property and some `uri`.
```
philbrookes: I agree with this sentiment, but am uncomfortable addressing this in the mobile-cli, I feel that this needs to be addresssed in the APB which defines it against whatever standard is agreed, and the mobile-cli could take steps to point out when fields that are required are missing.

i.e. if we decide that this field should be called `uri` then any that have called it `url` would get an error when interacted with via the mobile-cli stating: `the uri field is missing`. This will help standardise APBs, ease APB development, and reduce the amount of maintenance required in the mobile-cli.

wtrocki: +100, No matter where we will resolve this field should be exposed in standard way.
I think that this proposal we abstract from where json may come from. 
We could agree on what is needed (and if that is possible) and then see how to enforce/document this for apb and cli.

Mobile-configuration.md
```diff
@@ -0,0 +1,67 @@
...
+}
+```
+
+### Service object
+
+```json
+{
+  "id": "string",
+  "name": "string",
+  "type": "string",
```
philbrookes: @wtrocki as we discussed on bluejeans a moment ago, it would be good to add a:
```
"bindings": ["id", "id"]
```

field to the service object too.

wtrocki: +100 for that. In Json schema this will look like this:

```
"bindings": {
  "type": "array",
  "items": {
    "type":  { "$ref": "#/Service/id" },
   }
}
```

I'm going to wait for more opinions and add that

Mobile-configuration.md
```diff
@@ -0,0 +1,67 @@
...
+  "url": "string",
+  "metadata": "object"
```
philbrookes: I think metadata might be a misleading name, as the values in here are still necessary for use of the service in question, they are just non-standard data (e.g. `realm` in the keycloak config), so could we go with something like:
- "custom"
- "specific"
- "non_standard"

wtrocki: Fully agree. What we will do is to probably have metadata defined on the service level - so both SDK and APB will send and expect the same format. In terms of names I think anything that describe unstructured (and nested data) will work here. Metadata is definitely a bad choice. 
Let's get more options from others and pick one. I have no preference, but it also may be:

- "service-data"
- "service-config"
- "config"

Mobile-configuration.md
```diff
@@ -0,0 +1,67 @@
...
+## Proposed Solution
+
+Provide standard for mobile configuration that will be extendable and have backwards compatibility.
+
+### Mobile cli specifics
+
+Mobile-cli currently print configuration out to the std. 
+Mobile-cli should have option to create `mobile-services.json` file in specified location.
+
+### Json format specific
```
secondsun: @wtrocki Are you imagining these files being a single file or multiple?

Mobile-configuration.md
```diff
@@ -0,0 +1,67 @@
...
+
+```json
+{
+ "version" : "number",
+ "serverUrl" : "string",
+ "namespace": "string",
+ "services": { "type": "array", "items":"Service"}
+}
+```
+
```
secondsun: Might we want to think about client and integration configuration objects in the proposal as well?

I know that on the client side, there are some common configurations we are starting to identify.  We will probably want standardized headers to be attached to all requests, some form of consumable security policy, etc.

For integrations we could probably provide some data about the type of integration we are using.  IE is a service meant to be used as a proxy, do we need to decorate requests, does one service validate requests to another, etc.  (These ideas are all spitballs)

Mobile-configuration.md
wtrocki: This is single json schema definition. I just split that for readability. 
In JSON schema you can define separate objects and reference to them in other objects. 

So result of this proposal will be PR with single JsonSchema that will be contract between SDK and APB. Every time things will need to change we will need to edit schema and bump version. 
It's pretty standard way.

Mobile-configuration.md
wtrocki: I would love to iterate on those and bump versions but for the initial version the main target is to clear some mess and agree on things we know. We can iterate later on schema and bump versions that will allow all SDK's to adjust.

wtrocki: Added new info to make this clear. 

> For integrations we could probably provide some data about the type of integration we are using. IE is a service meant to be used as a proxy, do we need to decorate requests, does one service validate requests to another, etc. (These ideas are all spitballs)

See conversation above with proposition. If you ok with that I will add that to proposal.

Mobile-configuration.md
wtrocki: Added info.

Mobile-configuration.md
maleck13: This is also covered by the the proposal linked above. Where it specifies the keys that should be present in the configmap and what they should be called.

Mobile-configuration.md
maleck13: What is the purpose of adding the bindings? Not against it just curious. My imagined workflow is as follows:

1) developer provision sync 
2) get config
3) write code to do sync
Happiness
4) provision keycloak
5) integrate sync with keycloak
6) get new config
7) write code to login user and get token pass it to sync sdk
Happiness 

Wondering what the bindings does here? I couldn't imagine a situation where the developer is unaware that keycloak exists and that sync is now being protected.

Mobile-configuration.md
maleck13: I agree with iteration here. Lets get the first version of the config consumable and then iterate over time. Agreeing here doesn't mean it is what the final GA config will look like.

Mobile-configuration.md
wtrocki: SDK networking layer needs to know. This dependency has significant impact on SDK architecture because implies that sync will also need keycloak sdk dependency etc. I think this area needs special proposal with involvement from the Services 1 team. 

I was talking about that with @aidenkeating and it seems that Core SDK needs to be aware if sync, keycloak are joined together. Same will happen when API will be there. 

> write code to login user and get token pass it to sync sdk

This assumes that sync and keycloak have strict dependencies between each other and also users making this choice in code. The way android sdk was implememented is that services initialize themselves basing on config and things work out of the box (at least in theory)

Mobile-configuration.md
wtrocki: +1 for checking that on apb testing (if that's possible)

wtrocki:
@maleck13 @philbrookes - I think that we had really good conversation here and proposal seems to be in good place for initial version. Is it ok to get your review/approval?

We have conversations here on some additional extensions that will be introduced at later stage. As we agreed It's to early to have them in the proposal.

Mobile-configuration.md
maleck13: I am not involved enough with the SDK to really comment here, but it seems really strange to me that the core sdk needs to know if keycloak and sync are joined. 
@aidenkeating any insight here would be good. 
I would have expected that a token would be retrieved from the keycloak sdk and simply passed to via the exposed API to anything that needed it.

Mobile-configuration.md
```diff
@@ -0,0 +1,82 @@
...
+Service specific metadata prevents us from strict json parsing.
+
+## Expectations
+
+- Provide mobile configuration that has only esential fields for mobile users
+- Standarize format so users to not break functionalities once users will start using configuration
+ 
+## Proposed Solution
+
+Provide standard for mobile configuration that will be extendable and have backwards compatibility.
```
maleck13: backwards compatibility with what? We don't have any release yet :)

Mobile-configuration.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Mobile cli specifics
+
+Mobile-cli currently print configuration out to the std. 
+Mobile-cli should have option to create `mobile-services.json` file in specified location.
```
maleck13: I don't mind this, but wondering if really needed? ```oc plugin get sdkconfig <client> -o=json > mobile-services.json```

Mobile-configuration.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Contract between SDK and AeroGear Services Backend
+
+AeroGear services cli (and maybe UI) should clearly define format for the interaction with the SDK.
```
maleck13: AeroGear services cli (and maybe UI) should **have a** clearly defined format for interaction with the SDK

Mobile-configuration.md
```diff
@@ -0,0 +1,82 @@
...
+To do that efficiently we suggest to use JSONSchema format with versioning.
```
maleck13: 👍

Mobile-configuration.md
```diff
@@ -0,0 +1,82 @@
...
+
+Each backend service will need to conform to the schema in order to get supported.
```
maleck13: This really is covered by the other proposal as the cli actually does the work to turn the configmap into the config. Currently we expect a set of labels and a uri key

Mobile-configuration.md
wtrocki: This implies that users will need to know what should be the file name and know linux streams. 
Going to remove that but this will mean work work on documentation. 

I always prefer to do more code and less documentation :P

Mobile-configuration.md
wtrocki: It is more about avoiding if statements when parsing json files with unregulated format. Fixing wording

Mobile-configuration.md
maleck13: I think we can mark this as resolved, but can we add a link to the proposal here as it does seem to address the main issue outlined here

Mobile-configuration.md
```diff
@@ -0,0 +1,81 @@
...
+## Expectations
+
+- Provide mobile configuration that has only esential fields for mobile users
+- Standarize format so users to not break functionalities once users will start using configuration
+ 
+## Proposed Solution
+
+Provide standard for mobile configuration that will be extendable and will be easy to parse in mobile clients.
+
+### Mobile cli specifics
```
maleck13: I think we can just remove this as it is a detail about the cli implementation not the format of the configuration

Mobile-configuration.md
maleck13: So here I think it should be that the CLI should output in the correct format. So if the expected fields are not present (just URI currently) in the configmap then that services config will be skipped.

Mobile-configuration.md
```diff
@@ -0,0 +1,81 @@
...
+This schema will be supplied in additional `metadata` parameter.
+
+### Initial top level json schema
+
+#### Top level object
+
+```json
+{
+ "version" : "number",
+ "serverUrl" : "string",
```
maleck13: perhaps change to cluster name or identifier. Not sure exposing this to all clients is what end users would want

Mobile-configuration.md
maleck13: 👍  for config

maleck13 (changes requested):
Some minor changes but I think we are close to merging now

wtrocki:
@maleck13 I addressed all comments by applying your suggestions. Because this proposal is related directly with SDK I have created dedicated folder and moved all proposals from top level to `sdk` folder.  Hope that is ok.

maleck13 (approved):

maleck13:
@philbrookes going to go ahead and merge. Further updates and changes can be appended as they come up

wtrocki:
Thank You so much guys for involvement and shaping this together!

PR #18, opened 2018-01-25
Versioning

ccfoley:
A proposal looking at how system components inside AeroGear should be versioned, primarily focused on a Mobile Service and its sub-components.


versioning/versioning.md
```diff
@@ -0,0 +1,113 @@
...
+## Additional Considerations
+- the relationship between the Core SDK and the Service SDKs from a version perspective! This relationship will need to be handled but the developer more than likley will not see the Core SDK or not have to worry about it.
+- How much defensive functionality is developed around allowing or disallowing Mobile Service Integrations to be requested if we know that there will be compatiability issues.
+  - this could be addressed in the IDE Plugin or the Mobile CLI. It possibly doesn't need to be addressed now but can be considered in the future.
+- the following will most likely also need versions
+  - Mobile Core (installer?)
+     - the installer will point to the repo where the playbooks are installed, currently aerogear/playbooks
+     - in downstream, it may point at a different organisation
+  - MiniShift Plugin
+  - Mobile App Representation (their APBs) - do they have a dependency on a Mobile Service, possibly CI/CD service? Worth considering
```
ccfoley, line 113: @maleck13 Do you think there will be a need to version 'Mobile App REpresentations'? In that they are created from a Custom Resource Definition which is created by the Mobile-Core; do the Mobile App Representations just fall in under the version of Mobile-Core and we don't have to worry about them as individual components which need their own version?

maleck13: the mobile clients are a CRD so are versioned naturally as per the general kubernetes versioning system
I don't see there being an impact on mobile services, I think it will only be the mobile cli, IDE and UI that use these directly

versioning/versioning.md
```diff
@@ -0,0 +1,113 @@
...
+| UI                   | v1.0 |
+| CLI                  | v1.0 |
+| E2E Service Tests    | v1.0 |
+
+The reason for introducing this is two fold;
+* to encapsulate all of the sub components and their versions such that developers can easily identify what versions of the sub-components make up that version of the service and therefore work together.
+* to have some high level entity which has a version when considering Mobile Service integrations (i.e. what Mobile Services can work together and what versions of these services work together)
+
+This table of each overall version of the Mobile Serivce and its set of sub-components must be produced and made availble through the docs. This table of information is essential for developers.
+
```
camilamacedo86, line 64: IHMO the user should not care about the tags when run/install/setup the mobile service. The latest version should be the latest released one. 

Regards the internal versions(deps) the user should be able to check the versions used to build their projects, I mean what libs/dep/version each project/application/service is using by each component in the case be able to troubleshoot it. For example, if the customer is facing a bug when trying login via the client app which is using the keylock feature they should be able via the OCP get the version of keylock feature used and tell us which will allow we check if that bug is solved or not in future version in the same way that I understand which it is possible to get done with the SDKs already. 

Then, this matrix should be not required at all.

wtrocki: > Then, this matrix should be not required at all.

It will be required for SDK/Showcase unless we are going to include them in UI (if we do then we can avoid matrix and complexity around it)

danielpassos:
@ccfoley Is this still valid?

ccfoley:
@danielpassos We have moved on really from this proposal.

PR #20, opened 2018-01-30
WIP: docs hosting proposal

pwright:


docs.aerogear-site-hosting.md
```diff
@@ -0,0 +1,21 @@
...
+Hosting html (and associated files) for docs.aerogear.org is essential for communication with users. Need to automatically publish any changes to mobile-docs repo to docs.aerogear.org.
+
+## Expectations
+
+- If we take an approach used by https://github.com/debezium/debezium.github.io where asciidoc and html files are both in same repo (adoc in develop branch, html in master branch), we can anticipate the repo getting large over time with history of html changes mixed with history of adoc file changes.
+
+  
+## Proposed Solution
+
+Create a docs.aerogear.github.io repo and use it to host docs.aerogear.org html files.
```
wtrocki: Better to avoid this name as this is githubpages antipattern. 
Best way to do it is to just create any repo name and link that to domain you want like docs.aerogear.org. 
It's really simple and it can be done by just single file in repository. 

docs.aerogear.github.io is used only when you do not have custom domain and it do not support subdomains.

docs.aerogear-site-hosting.md
```diff
@@ -0,0 +1,21 @@
...
+
+Use a CI service (eg circleci) to run asciibinder and push the output to the docs.aerogear.github.io repo.
```
wtrocki: Worth to state that this will rebuild some specific branch? Master?
We really need more info about this. 

Circle CI PR build will differ from release build.
I think that CircleCi is quite cool for verifying individual PR's and we will need separate release chain that will happen on specific branches.

Problem is that rebuild of the webpage docs needs to be aligned with the changes that caused docs change.
It's best to have specific branch that will be updated and tagged for release.

We hit this hard in RainCatcher - some cutting edge changes were merged into docs without release available.
Basically all challenges for this were resolved in RainCatcher project (this aproach is very similar to what we had)

docs.aerogear-site-hosting.md
pwright: @wtrocki  good point, updated

docs.aerogear-site-hosting.md
pwright: @wtrocki I'm thinking master, and have updated doc with my expectations, that there won't be 'release' of mobile.next upstream, therefore there won't be a version branch or tag

matzew:
reads nice!

:+1:

wtrocki:
@finp 👍

PR #22, opened 2018-02-09
Mobile App Metrics and SDK version tracking proposal (2nd one)

aliok:
- Old one: https://github.com/aerogear/proposals/pull/14
- Gathering information from the related spikes and docs.
- Branch created in upstream repository instead of own repository to allow for contributions from others.
- JIRA: https://issues.jboss.org/browse/AEROGEAR-1897


aliok:
cc @david-martin @pb82 @darahayes @wtrocki 
quickly prepared the proposal PR. please have a look so that we can merge before Wed.

metrics/default-mobile-app-metrics.md
```diff
@@ -0,0 +1,139 @@
...
+
+A Metrics Service will be available from the Service Catalog, which can be provisioned by developers for use with their mobile applications.
+
+The following diagram illustrates the components:
+
+![service diagram](./default-mobile-metrics-system-landscape.png)
+
+### Mobile App Client
+
+A native, platform-specific SDK for targeting the Metrics Service will be available to developers.
```
wtrocki: Cordova Plugins, React Native plugins

metrics/default-mobile-app-metrics.md
```diff
@@ -0,0 +1,139 @@
...
+
+Upon installation into a project and providing configuration for targeting the Metrics Service the SDK will automatically 
+send the default metrics information for the application with a unique ID to the metrics service.
+This information will be sent to the service every time the application starts.  
+
+Please note that, the client side is not part of this proposal.
```
wtrocki: Let's remove that statement. Mobile metrics are for mobile apps. Considering that mobile part is mostly done I think we should just explain interface for mobile. There is no need for separate proposal for mobile and this may suggest that there will be one.

aliok:
Some questions:
- We provision a new Grafana instance here. Is it possible to reuse the existing Grafana instance, if it is provisioned by backend metrics service? Which actually makes me wonder if having 3 separate services would make sense: visualization, backend metrics, mobile app metrics. I see we figured out dashboard discovery here: https://issues.jboss.org/browse/AEROGEAR-1886. @darahayes any comments
- Regardless of reusing or not, can we not embed dashboard in Grafana but make it discovered by Grafana? @darahayes 
- What language/platform to implement the service? @david-martin

metrics/default-mobile-app-metrics.md
aliok: Can you write the whole sentence?
I couldn't really understand, sorry

wtrocki: IOS, Android and Cordova Plugins, React Native plugins will be..

metrics/default-mobile-app-metrics.md
```diff
@@ -0,0 +1,139 @@
...
+
+Receives `Content-Type: application/json` containing JSON with following schema:
+```
+{
+  "clientId": "deadbeef123"
+  "appId": "com.example.someApp"
+  "sdkVersion": "2.4.6"
+  "appVersion": "256",
+  "platform": "android",
+  "platformVersion": "27",
```
aliok: `platform` and `platformVersion` is new in this document, but it is part of the story in Trello card.

wtrocki: Thanks for flagging that. Going to add that to IOS. For ios I'm going to hardcode platform as 'ios"
Ping @danielpassos for Android.

In terms of implementation I think that can be as separate metrics class on mobile that will in 100% match this example json.

wtrocki: @aliok We also have startup timestamp metrics: https://github.com/aerogear/aerogear-ios-sdk/blob/master/modules/metrics/metrics/StartupMetrics.swift#L12

This was added as simple example, but it will give useful info.
This will add new field: `startupTimestamp` to indicate actual time when app was launched. 
Because application can be offline etc. taking server time is not going to work.

wtrocki: @aliok We could also send device type for both platforms if that's useful.

aliok: hmm, good point about the timestamp. actually, this wouldn't be possible with Prometheus, but it should be possible with Postgres only. I was thinking if we should trust the user about the timestamp, but we trust them with versions etc. anyway :)

Timestamp should be in UTC w/o any time zone information!

aliok: @david-martin I leave it to you if we should have device type or not.

metrics/default-mobile-app-metrics.md
```diff
@@ -0,0 +1,142 @@
...
+  "timestamp": "12223123123"
```
aliok: `timestamp` added.

metrics/default-mobile-app-metrics.md
```diff
@@ -0,0 +1,142 @@
...
+}
+```
+
+Timestamp must be in UTC, without any time zone. Daylight saving should also be discarded. Server side application will get the timestamp and convert it to timestamp with server's timezone.
+
```
aliok: I made the bold statements above about the timestamp, but would love to hear what you're going to say.

metrics/default-mobile-app-metrics.md
wtrocki: See https://github.com/aerogear/aerogear-ios-sdk/pull/15/files

pb82: @aliok @wtrocki Is the timestamp really useful? The timeseries data by definition is timestamped already and for non-timeseries data the database can easily add one to each record should we want it. No need to add it to the payload or do you see a good reason?

metrics/default-mobile-app-metrics.md
```diff
@@ -0,0 +1,142 @@
...
+A thin HTTP-based service will be present to receive metrics data from the mobile applications and then to forward to the storage service.
+
+At a PoC level, the endpoints available from the Mobile Metrics Service should be:
+
+`POST /default-metrics`
+
+Receives `Content-Type: application/json` containing JSON with following schema:
+```
+{
+  "clientId": "deadbeef123"
```
wtrocki: @david-martin @aliok @pb82 - Do we want to namespace different metrics? For example for IDM will we have additional object here or all should be merged into single object. 
```
"idmMetrics": {
 "rooted":true
}
```
vs
```
"idmMetricsRooted":  true
```

IMHO it will be best if this object will contain only metrics on the top level like:
```
{
  "versionMetrics":{..},
  "idmMetrics":{..},
  "platformMetics":{..}
}
```

That will be cleaner for SDK and will allow to use different storage engines in the future?

david-martin:
> We provision a new Grafana instance here. Is it possible to reuse the existing Grafana instance, if it is provisioned by backend metrics service? Which actually makes me wonder if having 3 separate services would make sense: visualization, backend metrics, mobile app metrics. I see we figured out dashboard discovery here: https://issues.jboss.org/browse/AEROGEAR-1886. @darahayes any comments

I think to a single APB that has the server metrics, mobile app metrics and visualisation is OK for an MVP. Grafana can be shared then. And Promethues can store data in Postgres.

> What language/platform to implement the service? @david-martin

What was used during your investigation?
Something relatively lightweight would be preferable.
Having a flexible Postgres library available in the language would be important (to allow storing both time series and structured data).
Language choice should also take into account downstream productisation concerns, so an already widely used language is preferable e.g. Golang, Node.js

wtrocki:
@david-martin +100 for golang implementation

metrics/default-mobile-app-metrics.md
aliok: The only scenario that it makes sense is offline. If the device was offline when the app was launched, we could record that and send to the server with the timestamp later.

I didn't really think about the effect on payload size. Even though it is a big thing, maybe it is better to make it optional. So, when the timestamp is missing, server pushes it with server time.
@wtrocki @pb82 what do you think guys?

@pb82  maybe I misunderstood your point.

metrics/default-mobile-app-metrics.md
aliok: @wtrocki I think what you're saying makes sense. but I leave this decision to @david-martin

aliok:
> I think to a single APB that has the server metrics, mobile app metrics and visualisation is OK for an MVP. Grafana can be shared then. And Promethues can store data in Postgres

Ok, sounds good to me. We can talk more about it during planning phase.

> What was used during your investigation?
> Something relatively lightweight would be preferable.
> Having a flexible Postgres library available in the language would be important (to allow storing both time series and structured data).
> Language choice should also take into account downstream productisation concerns, so an already widely used language is preferable e.g. Golang, Node.js

I used Node. Golang is also fine. Has the required libs.

metrics/default-mobile-app-metrics.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Problem Description
+
+As mobile apps are developed, improved and iterated on, different versions of those apps can be used on many different devices. 
+Metrics data becomes invaluable for developers to observe the landscape of application installations and make sure further 
+development and deployment of backend services will not affect users.
+
+## Proposed Solution
+
+A Metrics Service will be available from the Service Catalog, which can be provisioned by developers for use with their mobile applications.
```
josemigallas, line 29: Are we talking here about extending the existing Metrics or creating a new one for mobile?

metrics/default-mobile-app-metrics.md
josemigallas: I think the timestamp should only be used when offline, managed by the SDK. From the backend perspective, as @pb82 pointed out it isn't that useful since we can/should use the one from the DB (therefore we don't have to trust the user as @aliok said, again only when online).

pb82: @aliok @josemigallas No that was basically my point: only add fields to the payload if they are actually useful and keep it minimal otherwise. If the device is offline, how would it be able to send metrics data? At the moment there is no sync/caching for metrics, the request would just fail. Maybe that's something we should look into in the future.

david-martin: > At the moment there is no sync/caching for metrics, the request would just fail. Maybe that's something we should look into in the future.

@pb82 agreed on this. It needs some thought. (It could be solved in the metrics sdk. However, sync is meant to solve the offline problem, so that could be part of the solution).
@johnfriz this may be something for the backlog

josemigallas: Well I was thinking the metrics SDK would just store those metrics while the device is disconnected and when connection is back, send them to the metrics endpoint anyhow.

metrics/default-mobile-app-metrics.md
aliok: For MVP, we're talking about bundling them into single APB.

josemigallas: I'm obviously too late for this conversation but if we plan to have 2 different services, wouldn't it be simpler/quicker to have them separated from the start?

aliok:
Just a note for future reference:

There is a bad thing about using TimescaleDB/Postgres for metrics: flexibility.
In ElasticSearch, we can just push JSON without defining anything on the storage system. 
But in Postgres, for example, if we want to push a new metric called "osVersion", we need to update the SQL table schema (`ALTER TABLE bla`).

I think this is OK for default metrics (appVersion, sdkVersion, etc.) that we're interested in this epic, but in the future when we want to provide user-defined metrics to our users, we can't just go and create/alter tables.

In case of using Prometheus (architecture#4), Prometheus was creating/altering tables for us. But they have created a super generic non-normalized table structure that can hold anything even though there's a performance cost. Maybe we can have a look how they structured tables (I think there was only 4 tables created by Prometheus: labels, values, etc.) later when we're working with custom metrics.

So, summary:
* for now, we go with normalized tables (columns per metric: "sdkVersion", "appVersion", ...). [Note that, this structure wouldn't make sense if the metrics weren't received at the same time]
* in the future, we need to create a separate table structure for custom metrics.

metrics/default-mobile-app-metrics.md
```diff
@@ -0,0 +1,142 @@
...
+
+| Time series data                                          | Regular data                           |
+|-----------------------------------------------------------|----------------------------------------|
+| You normally do inserts                                   | You do inserts, but updates more often |
+| Getting some stats about what happened in the last N days | Knowledge base                         |
+
+Typically, what is tried to achieve here is done with inserting time series data in a time series database.
+This answers the question "what's the number of SDK init events for SDK version 1.2.3 in last N days?"
+
+However, in order to answer the question "what's the usage of SDK version 1.2.3 among clients?" we need to eliminate the duplicate events per client.
```
wei-lee, line 118: Just curious, is this the only reason that we choose to implement our own solution (a new service + storage)? If that is the case, why not we just avoid duplication from client sdks? We can relatively easy to  make sure each client only send in the metrics data once?

wei-lee: What I am looking at is that if we can make sure each client only send the metric data once, then perhaps we can just use Prometheus as the backend service?

aliok: @wei-lee I think you mean something like send metrics only on app install or update. That would eliminate duplicate data on client side directly.
This is something we talked and it is problematic. But even when assuming we solved this, we would still need "update" operations which Prometheus doesn't support.

> is this the only reason that we choose to implement our own solution

Yes, it basically is.

We're open to new ideas. Let me know if I misunderstood your point.

wtrocki: @wei-lee Really good question. We had this discussion as part of the https://github.com/aerogear/aerogear-ios-sdk/pull/13 

There are technical means to just provide information in SDK when version was changed and do not save duplicates.

darahayes:
@aliok You've done a lot more research here but would it be possible to do something like having a Postgres/Timescale table with a json/jsonb column? e.g:

```sql
CREATE TABLE usermetrics (
      time timestamptz,
      data jsonb
);
```

now we could insert like:

```sql
INSERT INTO METRICS VALUES (NOW(), {"sdkVersion": "1.1", "osVersion": 7, ...,}
```

Or perhaps we could do something generic like this:

```sql
CREATE TABLE usermetrics (
      time timestamptz,
      key varchar,
	  val varchar,
);
```
and we could query for a particular metric like this:

```sql
SELECT time, val FROM usermetrics WHERE key='sdkVersion';
```

obviously the examples here are trivial but you should get the point. I think option 1 gives lots of flexibility for inserts whereas option 2 makes it very easy to query. Would one of these approaches be feasible? Food for thought.

aliok:
@darahayes 
Haven't really experimented with "jsonb" column but I think it would bring some limitations in terms of indexing. Something we can have a look in the future though.

2nd idea seems nicer to me. But, homogeneous data is nicer when it comes to performance and it is best practice. Of course, we can sacrifice some things for flexibility. We can have a look at how Prometheus is storing data in TimescaleDB. I think it would give us a good idea.
Anyway, this is low priority for now IMO.

darahayes:
@aliok I have read that one advantage of the jsonb type is that you can do indexes. (tradeoff however is that inserts are quite slow)

Regardless, I agree it's low priority for now and we should go with the simplest approach for MVP. It's just good to know we have options later on!

metrics/default-mobile-app-metrics.md
wei-lee: > This is something we talked and it is problematic.

@aliok can you explain a little bit more on this?

> we would still need "update" operations which Prometheus doesn't support.

@aliok can you explain why we need the `update` operation?

aliok: > @aliok can you explain why we need the update operation?

This is in order to reduce the duplicates.

Imagine you have these init events:

t0 -> time
1.2.3 -> sdk version

1. clientA, t0, 1.0.0
2. clientB, t1, 1.0.0
3. clientA, t2, 2.0.0
4. clientC, t3, 2.0.0
5. clientD, t4, 3.0.0

Desired information is:
- 1.0.0: 1 client
- 2.0.0: 2 client
- 3.0.0: 1 client

Event#1 should not have an effect because clientA doesn't use 1.0.0 anymore.

In our case, we can't simply insert all of these events to come up with the desired information by using aggregation etc. (see https://stackoverflow.com/questions/48378449/elastic-search-get-latest-for-a-value)
So, we need to maintain a <clientId,sdkVersion> map to come up with the desired information. And, maintaining this map needs `update` operation.

>>This is something we talked and it is problematic.
> @aliok can you explain a little bit more on this?

This is some extra work, but possible. However, even when we assume have done this, we still come to the problem I explained above.

wtrocki: > This is something we talked and it is problematic. 

Actually in terms of the SDK that will be desired implementation, but I got impression that we want little bit more than just a version and we should send this information every time application starts.

Metrics for IOS SDK is finished and merged to master so I could do show and tell how this is implemented,.

aliok: We anyway need to send stuff on every launch, because of usage (event count) statistics. But it could've been a separate endpoint or something.

wtrocki: @aliok @wei-lee Considering number of really important comments I think it will be the best to have quick show and tell about Android and IOS implementations and how server side can look like.
WDYT?

wei-lee: Thanks @aliok. Although I don't think `update` operation is a good idea, as we lose information by doing it, I can understand for the approach we are going to take, it is probably the best option.

However, I am a little bit concerned about the approach we are taking. A few reasons:

1.  To me, it feels like we are building another version of `fh-messaging`/`fh-metrics`. Yes, I know we are only building a "simple" metrics service, but it is exactly how `fh-messsaing`/`fh-metrics` get started. We have to be careful not making the same mistake again.
2. The current proposed solution is a little bit "too simple": it is tailored only for the current goals and it doesn't extend.  This means if we need to add any new feature without turning this service into another `fh-messaging`/`fh-metrics`, we probably need to look at something else. This isn't necessary a problem, but it should be highlighted as a risk of choosing this approach.   

Based on that, I think we should add a section to describe the limitation of choosing this approach.

aliok:
@darahayes 
I was trying to keep the scope to default app metrics (app version, sdk version, etc.) for this proposal and it made sense to a structured table for those metrics. 

But I noticed IDM self defence metrics is coming soon and we will have more metrics.
I am now thinking instead of creating a structured table, we can use semi-structured for everything. This unifies the approach for metrics (default app metrics, IDM metrics, custom metrics, etc.).
I am gonna spike around that. The reason that I have noticed this problem very late is because we always had semi-structured tables/indices in previous approaches (ElasticSearch or Posgres over Prometheus).

metrics/default-mobile-app-metrics.md
aliok: @david-martin any comments on this?

david-martin: @josemigallas I think getting the mvp with a single APB for both service & mobile app metrics is OK. If there's a push/requirement to separate them into individual APBs, we can address that then.

wtrocki:
@aliok - I kinda feel that we will need to provide abstraction for metrics on the server and just assign data engines to map to actual storage system. Data engines may be something we will need to code to provide flexibility and adjust this over the time so we will not be tied to actual format of the data etc.

From SDK point of view we will have metrics containers (collectables) that will encapsulate specific fields.
On server this abstraction need to still exist (see my suggestion above) and map to specific storage requirement and presentation requirements (grafana dashboards). Having it this way we could document entire metrics extension process and make future contributions really easy.

metrics/default-mobile-app-metrics.md
wtrocki: @david-martin That's really important aspect that will help us to architecture this metrics service towards better SDK interpretability and future extensibility. It will save a lots of time for future metrics work and make it more community friendly. It's really important to get this into the layout as it will require minimal efforts but it will give us a lots of advantages in really short time period.

Do you have anything against this aproach?

david-martin: How about something like below (based on what @wtrocki suggested?

```json
{
  "app": {
    "id": "com.example.someApp",
    "version": "1.0.1"
  },
  "device": {
   "id": "deadbeef123",
   "platform": "android",
   "version": "27"
  },
  "idm": {
   "rooted":true
  }
}
```

aliok:
@darahayes @wtrocki I've checked how Prometheus done it. They use JSONB :)

So, I have created another module in the POC to do stuff with JSONB: https://github.com/aliok/mobile-analytics-poc/pull/6

This way, we only have one table with 3 fields in it : clientId, timestamp, data.
Field data is JSONB. This unifies the default metrics with custom metrics. In the POC, I have used the same table for sdkVersion metrics and a custom "button click" metrics.

I haven't really checked the indexing techniques though. As @darahayes pointed out, there are ways to index fields inside JSONB data.

@wtrocki I totally agree on what you are saying. We should do that!

metrics/default-mobile-app-metrics.md
aliok: @wei-lee There will be table that we are going to maintain because of the reasons above, but for the data that doesn't need any de-duplication, we will use plain time series database entries. We will just insert them and let the user aggregate data in the visualization layer.

For example, even though it is not really in the scope of this proposal, custom metrics like "how many times this button is pressed?" is something users won't want to de-duplicate. So, no table maintaining for that. ...simply event streams insterted in a table.

metrics/default-mobile-app-metrics.md
darahayes: Regarding the question of 'who creates the timestamp' I think the question 'What does the timestamp mean/represent?' helps us understand.

I would argue that the timestamp means 'the time at which the data was recorded.' To me that would mean it's added on the server side. However, it's not hugely important.

We could of course allow the client to add the timestamp to the payload and if it's not present then the server can simply add it. Just a suggestion.

metrics/default-mobile-app-metrics.md
david-martin: > To me, it feels like we are building another version of fh-messaging/fh-metrics. Yes, I know we are only building a "simple" metrics service, but it is exactly how fh-messsaing/fh-metrics get started. We have to be careful not making the same mistake again.

@wei-lee This is definitely not another version of fh-messaging/fh-metrics.
There is no rolling up of metrics (like what fh-metrics did).
This is done at Grafana query time. It allows much greater flexibility than fh-metrics.

I'd argue the main reason for a custom server is because Apps cannot talk directly to Prometheus. Also, there will need to be some element of security (api keys?) eventually, so an intermediary service between Apps and the underlying data store makes sense. Also, it would be necessary to validate the data coming from (untrusted) apps before storing the data.

> The current proposed solution is a little bit "too simple": it is tailored only for the current goals and it doesn't extend. This means if we need to add any new feature without turning this service into another fh-messaging/fh-metrics, we probably need to look at something else. This isn't necessary a problem, but it should be highlighted as a risk of choosing this approach.

That's a fair point that its tailored for the current goals.
I'm OK with that. It's not so complex that it couldn't be modified as more requirements are prioritised. (cc @johnfriz if you have any concerns here).
Again, I don't think it's fair to compare to fh-messaging/fh-metrics. That is a different pattern/paradigm with rolling up of data with cron tasks, and no security, apart from the core java app (millicore) sitting in front of it.
I think there needs to be a balance between flexibility and locking down what metrics can be stored. I think its too early to say that balance is off, given that the server hasn't been written.

wei-lee: @david-martin Yes, this is not another version of fh-messaging/fh-metrics yet, but it has the *potential* to become one. If we don't monitor it closely, I think it will be relatively quick to see more features like aggregation/cron jobs being added to this service. 

All I am saying is we just need to be careful to not let this service to grow become another fh-messaging/fh-metrics in the future. If there is any sign that this service is turning into something is that like a BI/analytics engine system, we should probably stop and ask if it is the right thing to do.

aliok:
@david-martin @wtrocki @pb82 @darahayes 
Made the changes in the proposal we talked today.
Do you mind giving another review to it?

metrics/default-mobile-app-metrics.md
david-martin: Thanks @wei-lee. To put it simply, this is a way to satisfy the relevant cards on the trello board, without investing much time right now, and drawing a line under the 'Metrics' work. A BI/Analytics engine/system is a completely separate set of requirements. That should be given appropriate time to understand what the requirements are and what's available if/when it is prioritised.

david-martin (approved):

david-martin:
@aliok Good to merge if there's another +1/approval.

wtrocki (approved):
All my concerns were addressed.
Really good work!

PR #23, opened 2018-02-13
Update Android SDK naming

aidenkeating:
@wei-lee Mind taking a look?

### DIFF
--- a/sdks/SDK-naming.md
+++ b/sdks/SDK-naming.md
@@ -13,10 +13,10 @@ Based on: https://groups.google.com/forum/#!topic/aerogear/pNmVTy0saDA
 
 ### Android:
 
-org.aerogear.android.ags.core - AeroGear Services Core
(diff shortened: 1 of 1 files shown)


wei-lee (approved):
LGTM.

wtrocki:
@aidenkeating - Added more people to PR for wider awareness

secondsun:
So in the early days of the Aerogear project we had a lot of confusion about "Android java" and "Java SE Java".  I feel like Java on Java SE is still the default and having android instead of mobile is valuable.

At the same time mobile + Java is pretty much Android these days, and as long as the service team isn't writing Java server services in org.aerogear.mobile then I'm not going to object.

wtrocki:
@secondsun This was updated because package names were not aligned. 
Do you think we could update package names in code and revert that at some point?
I do not know how much will be involved in terms of bom etc.

secondsun:
@wtrocki Ahhh.  Code wins.

PR #24, opened 2018-02-14
Added proposal for E2E testing

wojta:
- [x]  Proposal for used frameworks 
- [x]  Proposal for test infrastructure and on device tests

### DIFF
--- a/testing/Android_SDK_E2E_Testing.md
+++ b/testing/Android_SDK_E2E_Testing.md
@@ -1,0 +1,76 @@
+# **Android SDK E2E Testing Proposal**
+
+[[TOC]]
+
+## Test suite
+
+### Outline of Required Test Suite Features
+
+* The test suite must be easily readable to make it enable those that are not familiar with the test suite to get involved quickly in the maintenance of existing tests and the production of new tests.
+
+* The test suite must be easily maintainable to reduce the time spent on fixing bugs.
+
+* The chosen framework to build the test suite must have the ability to switch between the native application and mobile browser to account for logging into an application using redirect to and from a web form.
+
+### Frameworks Explored
+
+The following frameworks were assessed by creating some proof concept tests against the mobile-security-android-template.
+
+#### Espresso
+
+Espresso is one of the java based frameworks that is favoured and documented on the Android developers website [1].  It provides automatic synchronization of test actions meaning the developer is not required to manage wait times between view transitions.  The framework can be used to test both native applications and hybrid applications and can be written in an easy to read syntax.  An Espresso test recorder is also included in Android studio to help the developer write code quickly.  However, Espresso does not allow any redirects to other apps to happen while testing an apps instead using intents to assert that the correct call was made. 
+
+#### UiAutomator
+
+UiAutomator is another java based framework that is favoured and documented on the Android developers website [2].  This framework is used to build test suites that can perform actions on both user and system apps.  Like the Espresso framework, UiAutomator does not allow for switching between the user application and the browser, again favouring the use of intents to assert that the call was made. 
+
+#### Appium
+
+Appium is a cross platform testing tool that allows for testing of native and hybrid applications on multiple platforms such as iOS and Android.  Appium has a number of clients that can be used to write tests in languages such as NodeJS, Java, Python, Ruby, etc. 
+
+#### WD.js
+
+This client has many positives - there is a large number of locator strategies to target elements in views, the syntax of such tests can be written in a promise chain, and its possible to use mochas transfer promisness to make assertions a natural part of promise chain.  However, despite the positive features, the Wd.js client cannot handle a redirect to the browser at the present time - the problem seems to be that all chromedriver sessions are terminated when attempting to switch context from native to webview which cause the tests to crash. 
+
+#### Webdriverio (nodejs)
+
+Like the WD.js library like the WD.js client, tests using this client can be written in a short chained syntax and can also be used in conjunction with features of tools like mochas transfer promisness to make assertions part of this chained syntax.  Unlike the WD.js client however, webdriverio can handle the redirect to the browser from the application, however there is only two possible locator strategies - xpath and accessibilityId. 
+
+#### Java-Client (Appium Java SDK)
+
+Unlike the javascript libraries the java-client cannot be written in a promise chain syntax, however, there is a large number of locator strategies and it can handle the transition between native application and mobile browser.  Another major feature of the java client is the support for page object classes - using the following annotations, @FindBy(someStrategy), @AndroidFindBy(someStrategy) and @iOSFindBy(someStrategy), allows the creation of a single page object when creating tests for applications on multiple platforms that have the same content and flow but identifiers unique to their platform rather than a separate page object for each [3]. 
+
+### Framework Recommendation
+
+Appium is the framework of choice from those explored as it is the only framework explored that provides clients that can handle the redirect to the browser that is an integral test case we must cover.  Of the two clients, the Java-Client is likely to be the superior choice as it has a greater number of locator strategies and the annotation selector targeting provided can reduce the size of the codebase.
+
+### Test suite Proposal
+
+#### Implementation Language
+
+With the selection of the Appium framework and its Java-Client java, or a JVM based language such as Kotlin, Groovy, Scala, etc. should be used, provided they can support the java based annotations that the appium Java-Client provides for Page Object support.  The use of a JVM language other than Java could increase the readability of the test suite.
+
+#### Page Objects
+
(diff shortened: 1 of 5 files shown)


testing/Android_SDK_E2E_Test suite.md
```diff
@@ -0,0 +1,71 @@
...
+
+With the selection of the Appium framework and its Java-Client java, or a JVM based language such as Kotlin, Groovy, Scala, etc. should be used, provided they can support the java based annotations that the appium Java-Client provides for Page Object support.  The use of a JVM language other than Java could increase the readability of the test suite.
+
+### Page Objects
+
+The test suite should make use of the page object model. This model separates the logic the presentation logic of the associated page objects and the test logic which leads to tests that are less likely to be brittle, and would therefore increase the maintainability of the test suite [4].  As mentioned in the framework recommendation section, the ability of the Appium library to use annotations for targeting selectors that are unique to each platform can help to reduce the codebase size - only one page object is required per platform with the annotations providing the ability to target Android or iOS specific selectors and the tests files themselves can also be generic meaning no separate tests for Android or iOS applications that share the exact same pattern of logic flow through the application.
```
wojta: Just note that if possible we should focus on having similar "example" applications for iOS and Android.

testing/Android_SDK_E2E_Test suite.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Java-Client (Appium Java SDK)
+
+Unlike the javascript libraries the java-client cannot be written in a promise chain syntax, however, there is a large number of locator strategies and it can handle the transition between native application and mobile browser.  Another major feature of the java client is the support for page object classes - using the following annotations, @FindBy(someStrategy), @AndroidFindBy(someStrategy) and @iOSFindBy(someStrategy), allows the creation of a single page object when creating tests for applications on multiple platforms that have the same content and flow but identifiers unique to their platform rather than a separate page object for each [3]. 
+
+## Framework Recommendation
+
+Appium is the framework of choice from those explored as it is the only framework explored that provides clients that can handle the redirect to the browser that is an integral test case we must cover.  Of the two clients, the Java-Client is likely to be the superior choice as it has a greater number of locator strategies and the annotation selector targeting provided can reduce the size of the codebase.
```
wtrocki: It will be good to create ticket for IOS Swift support. 
More context:
https://discuss.appium.io/t/does-appium-support-swift-language/7980/3

wtrocki:
@wojta We have template for proposals: https://github.com/aerogear/proposals/blob/master/template.md

I think that main risk for this is that at some point it may happen that examples may diverge because of platform specifics and we will not be able to use single test suite. Making sure that both example apps will work with the same test suite will require extra effort etc.

Another smaller risk is IOS, Cordova, React, JS and Xamarin support. 
It will be nice to get some separate sections for explanation and point out what platforms we going to cover here.

testing/Android_SDK_E2E_Test suite.md
wtrocki: IMHO - that may too low level for proposal. 
Initially we just need to know - what is going to be tested, what tooling will be created etc. 
It will be good to present just top level overview.

wtrocki:
@wojta There are two ways to make this proposal mergeable:
A focus on Android only
B extend proposal to cover other platforms

Do you have any code/examples/investigations that can be seen for this proposal.

testing/Android_SDK_E2E_Test suite.md
wojta: Actually we don't need to have it in Swift, it's client side language only, Appium Java SDK can test both Android and iOS.

wojta:
@wtrocki  
https://github.com/wojta/mobile-security-android-template/blob/8d4742dcb596e637c11f88c63114016cd1322263/integration_tests/src/test/kotlin/com/feedhenry/securenativeandroidtemplate/integrationtests/LoginTest.kt

https://github.com/twnolan/appium-experimentation/pull/2

We don't have demo for page objects in Java/Kotlin yet.

wtrocki:
@wojta that looks really really promissing. It may be worth to add that example to proposal.

testing/Android_SDK_E2E_Test suite.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Page Objects
+
+The test suite should make use of the page object model. This model separates the logic the presentation logic of the associated page objects and the test logic which leads to tests that are less likely to be brittle, and would therefore increase the maintainability of the test suite [4].  As mentioned in the framework recommendation section, the ability of the Appium library to use annotations for targeting selectors that are unique to each platform can help to reduce the codebase size - only one page object is required per platform with the annotations providing the ability to target Android or iOS specific selectors and the tests files themselves can also be generic meaning no separate tests for Android or iOS applications that share the exact same pattern of logic flow through the application.
+
+### Self-Contained Repository
```
wei-lee: So if the tests are in a separate repo, when they will be triggered to run? How can SDK developers get quick feedbacks if something  is broken?

wtrocki:
@wojta As agreed I will try out apium on IOS to unblock and provide feedback for our example app.

testing/Android_SDK_E2E_Test suite.md
twnolan: @wei-lee What Appium requires is a path to the produced app file so it can upload it to the emulator and test it. This path is specified within the options passed to the appium client when running a test.

On a developer machine we could use a file the the location of build directories of multiple SDKs which will provide this information to the appium configuration options section. The steps to run the e2e tests then would be:

A. Trigger build from SDK repo
B. Trigger UI test command from e2e testing repo

Of course a shell script or system alias could be used to trigger these commands sequentially without the developer needing to manually trigger them both.

testing/Android_SDK_E2E_Test suite.md
twnolan: @wtrocki 

> I think that main risk for this is that at some point it may happen that examples may diverge because of platform specifics and we will not be able to use single test suite. Making sure that both example apps will work with the same test suite will require extra effort etc.

We can still create tests specific to each as well, but for anything that is common we could have common tests. There is a possibility to write our tests to cover multiple os platforms which we are highlighting here.

There should also be no extra effort to make the apps work from the app development side, but I am assuming there is some sort of planned common UI flow logic and layout. The specifics of things like ids, classes, etc. do not matter.

> Another smaller risk is IOS, Cordova, React, JS and Xamarin support.

There is no problem in support. Appium can test native, hybrid and mobile web apps on iOS, Android, and Windows mobile devices. A quick web search will show tutorials of how these are achieved in the technologies you have suggested.

> IMHO - that may too low level for proposal.
Initially we just need to know - what is going to be tested, what tooling will be created etc.
It will be good to present just top level overview.

In my view a high level is fine if you are starting from scratch, but there has been a lot of investigation into frameworks done already. With this information the proposal should then outline the pros and cons of each framework (done), and produce a reccomendation (done).

While we started out on considering the Android first, Appium can test any mobile apps built from the SDKs we plan to create and more. Buit Android is our starting point.

When writing the proposal for e2e testing this recommendation is how I have interpreted the framework/test suite to be created from investigations we undertook, discussions with @wojta and adhering to best practices like page objects. Specific details of how page objects can be created are there to illustrate the possibilities in reducing codebase size and increasing maintainability but may end up not being used in the actual implemented code.

testing/Android_SDK_E2E_Test suite.md
twnolan: Any Appium Client can be used to test native, hybrid and mobile web apps on iOS, Android, and Windows mobile devices. A full list of clients available can be found here - https://appium.io/docs/en/about-appium/appium-clients/

wei-lee:
In principle I think worth giving Appium a try. 

For me, I think choosing which framework to use or writing the tests aren't the biggest problem, but setting up the right testing infrastructure and making the E2E tests automated is. I would like to see if it is possible to get some POC done to prove that the proposed solution will work.

Also would like to see the feedbacks from the QE team. @AdamSaleh  @psturc do you guys have any thoughts/ideas/comments?

wojta:
I have vitrified proposals and put them into the new files and changed them according to the template. Old ones are still there in separate directory, because they contain relavant information about what led to the final proposal.

[E2E_Test_suite.md](https://github.com/aerogear/proposals/pull/24/files#diff-18dba767f14e8055360b6fe461dfcc0b)
[E2E_Test_infrastructure.md](https://github.com/aerogear/proposals/pull/24/files#diff-c1b95f088919919e49f66f7e2a423ccd)

@twnolan please recheck if it's correct.
@wei-lee @wtrocki @AdamSaleh @josemigallas if you can comment and review

psturc (approved):

psturc (approved):
Great work Vojta, really good and detailed explanation of what we want to achieve in e2e testing for 5.x.
I have not much experience with the suggested software and libraries, but overall the proposed solution makes sense to me.

wojta:
Thanks @psturc. Great work of @twnolan also.

AdamSaleh (approved):

twnolan:
@wojta This looks fine to me.

PR #25, opened 2018-02-16
JS (Cordova, React Native & Web) proposal

wtrocki:
Early version of AeroGear Services Cordova proposal.
Target of this proposal is to provide overview for Cordova based support, but it may also cover Web and ReactNative support. I have added list of the TODO's to proposal. 
Any suggestions welcomed.

See also community discussion: https://groups.google.com/forum/#!topic/aerogear/FGfpBdNxe-4


sdks/Cordova.md
```diff
@@ -0,0 +1,138 @@
+# Apache Cordova AeroGear Services
+
+## Motivation
+
+Provide set of the SDK's that can be used for Cordova based development.
+Provide best performance and user experience on the platform.
+
+## Proposed solution 1
+
+Pure javascript solution for the mobile configuration and orchiestraction
```
pb82: typo: orchiestraction -> orchestration

darahayes:
I will say +1 to pure JS implementation of the 'core'/config parsing part. It should be possible to deliver the networking and config parsing parts in JS and have it compatible across web/cordova/React Native.

However, I don't think there will be one approach that works for all SDKs. I think we will need to evaluate what the best option is (Pure JS, Native wrappers, etc) on a case by case basis for the service SDKs.

sdks/Cordova.md
```diff
@@ -0,0 +1,138 @@
...
+## Motivation
+
+Provide set of the SDK's that can be used for Cordova based development.
+Provide best performance and user experience on the platform.
+
+## Proposed solution 1
+
+Pure javascript solution for the mobile configuration and orchestration 
+that will allow cooperability between JavaScript and Native.
+Mixing both existing Native plugins and pure JavaScript based solutions was identified as the most reasonable aproach.
```
wei-lee: I think this line is still a bit confusing for me. Does it mean we always have a native & js implementation, or we should look at the possibility of having a pure JS implementation first?

I think it needs wording a little bit better to avoid any confusion.

sdks/Cordova.md
```diff
@@ -0,0 +1,138 @@
...
+Configuration parser and all AeroGear Services specific business code doesn't require any native implementation and can be reused in web/cordova/react native implementations. Individual SDK will may by default be implemented in pure JavaScript or relay on the cordova plugins.
```
wei-lee: I am not sure I get the first sentence.

sdks/Cordova.md
```diff
@@ -0,0 +1,138 @@
...
+- Aligned with the way cordova works (single plugin to do specific *native* job, integrated with others in JavaScript)
+- Ability to reuse existing cordova plugins that perform native tasks (push notifications, sync, auth)
+- Ability to wrap existing JavaScript solutions and use them without 
+- Ability to select and use only elements that are needed in the app.
+
+See investigation details bellow for more details
+
+## Proposed solution 2
+
+Pure JavaScript implementations (JavaScript SDK) that will abstract from browser support.
```
wei-lee: It has both "implementations" and "abstract" and I am not sure what this mean

wei-lee:
I would love to give a "+1" or "-1" on this proposal, but at the moment, I just don't know if I understand it. 

What is the biggest difference between proposal 1 & 2? I didn't quite get it.

sdks/Cordova.md
wtrocki: Thank you. I will correct that to make sure that we mean JS first and plugin when it is not possible/wrapper is needed.

sdks/Cordova.md
wtrocki: I means that javascript code for example will not assume that `document` variable is available and provide wrapper for ReactNative/Cordova etc. Really sorry for bad wording. Going to explain that carefully.

wtrocki:
TL;DR - I think the best way to summarize

1) Write mixture of JavaScript and Cordova/React plugins. This path do not support web - it's just separate line of work. More Cordova/React Native specific. (Example: https://github.com/aerogear/aerogear-cordova-push)
2) Should we write it to the Web first in pure JavaScript and provide alternative methods to execute code for Cordova and ReactNative. (Example: https://www.npmjs.com/package/keycloak-js ) 

Which one will be your choice? I'm leaning more towards no 2 at the moment.

sdks/Cordova.md
pb82: I think the idea is that we write the Cordova SDK in Javascript and use abstractions for APIs that are browser specific (e.g. `fetch` or `axios` for http requests).

wtrocki: @pb82 Good explanation. Thank you for that!

wtrocki:
@pb82 This proposal was initially only about Cordova plugins but after applying some suggestions from aerogear list scope was extended - solutions that were proposed expanded to Web and ReactNative.
I would love to see your point of view on this. Going to leave this proposal open and focus on actual implementations, so we could extend it once you will be happy with ReactNative spikes.

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+- Ability to reuse existing cordova plugins that perform native tasks (push notifications, sync, auth)
+- Ability to wrap existing JavaScript solutions and use them without 
+- Ability to select and use only elements that are needed in the app.
+
+See investigation details bellow for more details
+
+## Proposed solution 2
+
+Pure JavaScript implementations (JavaScript SDK) that will we written for the web and abstract from browser support.
+This means that specific browser functionalities that may not be available for React or Cordova will be provided to enable support for this platforms. JS SDK will be working with Web, Cordova and ReactNative - each platform will provide wrappers for specific apis if needed. This aproach will not allow us to reuse native implementations, but it will make Cordova, Web and ReactNative SDK's more consistent. Implementation could detect platform and load specific ReactNative/Cordova/Browser wrappers.
```
evanshortiss: I like this idea if I understand 👍 

So you mean something like this?

```js
const platform = require('src/platform')

if (platform.isReactNative()) {
  // Do RN implementation
} else if (platform.isCordova()) {
  // Do Cordova implementation
} else {
  // Web/Electron etc...
}
```

It would make the bundle a little larger to load, but also has all those advantages, and easier documentation due to a unified API.

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+Since some of the plugins like push are already in separate repository this can be seen as additional effort. 
+It may also complicate distribution as community preffers to have 1 plugin per repository and cordova cmd is optimized.
+I do not have strong opinions about this one but it feels like leaving things separate and having some aggregator repository with example application that links to all plugins and have parser code will be the best way.
+
+## Architecture for solution 1
+
+### Config parsing 
+
+Cordova config plugin not needed due to trivial implementation that can be done in pure JavaScript.
+Networking/Logging should be done on JavaScript side as this is common aproach for cordova.
```
evanshortiss: I'm playing devil's advocate here, but could this be a security issue, e.g storing logs securely for upload later? The commonly used storage mechanisms in Cordova and React Native are not encrypted AFAIK. Maybe this is a goal for later?

sdks/Cordova.md
wtrocki: @evanshortiss - You are right. I'm investigating now using webpack that will allow us to have single codebase that will produce different flavors with different code. Your code snipped will still work fine but we could have that if on build time and produce separate artifacts for the different platforms.

sdks/Cordova.md
wtrocki: Both react and cordova plug into `console.log` by plugin available by default. We could swap override plugin if anything specific will be needed.

sdks/Cordova.md
evanshortiss: That'd be much better since it'd have smaller JS bundle size 👍

evanshortiss:
One thing missing here is if we'd use Babel/TypeScript/CoffeeScript etc. Is this something to consider @wtrocki?

wtrocki:
@evanshortiss Good point. I'm workin on POC with TypeScript.

david-martin:
> One thing missing here is if we'd use Babel/TypeScript/CoffeeScript etc. Is this something to consider @wtrocki?

@evanshortiss @wtrocki 
In terms of implementation language, I don't think that matters from the mobile developer's point of view, as they will be consuming the sdk interface.
It does matter from a maintenance point of view, and should probably be decided by the commuinity who will be doing the implementation and maintenance. If TypeScript was decided, great! Good to call it out here in case there's any discussion to be had

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+## Motivation
+
+Provide set of the SDK's that can be used for Cordova based development.
+Provide best performance and user experience on the platform.
+
+## Proposed solution 1
+
+Pure javascript solution for the mobile configuration and orchestration 
+that will allow cooperability between JavaScript and Native.
+Configuration parser and other pure JS code willnot require any native implementation and can be reused for cordova/react native implementations. Individual SDK will may by default be implemented in pure JavaScript or as Cordova/React plugins.
```
david-martin: 'willnot' > 'will not'

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+Pure javascript solution for the mobile configuration and orchestration 
+that will allow cooperability between JavaScript and Native.
+Configuration parser and other pure JS code willnot require any native implementation and can be reused for cordova/react native implementations. Individual SDK will may by default be implemented in pure JavaScript or as Cordova/React plugins. 
+
+## Advantages of the solution 1
+
+- Significantly reduced effort to develop and maintain Cordova Support by reusing available components.
+- Aligned with the way cordova works (single plugin to do specific *native* job, integrated with others in JavaScript)
+- Ability to reuse existing cordova plugins that perform native tasks (push notifications, sync, auth)
+- Ability to wrap existing JavaScript solutions and use them without
```
david-martin: Unfinished sentence? 'use them without'  use them without what?

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+Provide set of the SDK's that can be used for Cordova based development.
+Provide best performance and user experience on the platform.
+
+## Proposed solution 1
+
+Pure javascript solution for the mobile configuration and orchestration 
+that will allow cooperability between JavaScript and Native.
+Configuration parser and other pure JS code willnot require any native implementation and can be reused for cordova/react native implementations. Individual SDK will may by default be implemented in pure JavaScript or as Cordova/React plugins. 
+
+## Advantages of the solution 1
```
david-martin: > Significantly reduced effort to develop and maintain Cordova Support by reusing available components.
Significantly reduced compared to what?
Also what do you mean by 'reusing available components'? What components are being reused? 
Would that be the same as the point about 'Ability to reuse existing cordova plugins that perform native tasks'?

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+
+- Significantly reduced effort to develop and maintain Cordova Support by reusing available components.
+- Aligned with the way cordova works (single plugin to do specific *native* job, integrated with others in JavaScript)
+- Ability to reuse existing cordova plugins that perform native tasks (push notifications, sync, auth)
+- Ability to wrap existing JavaScript solutions and use them without 
+- Ability to select and use only elements that are needed in the app.
```
david-martin: I don't understand this, from a technical perspective. Ability to select what exactly?
Is this about loading/including of libraries? What's the alternative to this that means all elements are 'selected'?

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
+# Apache Cordova AeroGear Services
```
david-martin: Should this be changes to AeroGear JS SDK Proposal?

Then call out Web, Cordova and React Native as the targets for this SDK?

sdks/Cordova.md
david-martin: 'for this platforms' > 'for these platforms'

david-martin: 'ReactNative' > 'React Native'

david-martin: 'apis' > 'APIs'   'aproach' > 'approach'

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+- Aligned with the way cordova works (single plugin to do specific *native* job, integrated with others in JavaScript)
+- Ability to reuse existing cordova plugins that perform native tasks (push notifications, sync, auth)
+- Ability to wrap existing JavaScript solutions and use them without 
+- Ability to select and use only elements that are needed in the app.
+
+See investigation details bellow for more details
+
+## Proposed solution 2
+
+Pure JavaScript implementations (JavaScript SDK) that will we written for the web and abstract from browser support.
```
david-martin: @wtrocki I'm not getting the difference between solution 1 & 2.

1 says `Pure javascript solution for the mobile configuration and orchestration ..... Individual SDK will may by default be implemented in pure JavaScript or as Cordova/React plugins`

2 says `Pure JavaScript implementations ..... each platform will provide wrappers for specific apis if needed`

They sound the same to me.

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+This means that specific browser functionalities that may not be available for React or Cordova will be provided to enable support for this platforms. JS SDK will be working with Web, Cordova and ReactNative - each platform will provide wrappers for specific apis if needed. This aproach will not allow us to reuse native implementations, but it will make Cordova, Web and ReactNative SDK's more consistent. Implementation could detect platform and load specific ReactNative/Cordova/Browser wrappers.
+
+## Advantages of the solution 2
+
+- Ability to collect and link 3 different platfroms as part of the single repository
```
david-martin: What does `collect and link` mean?

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+- Code still can be reused thanks to keycloak, push, sync js implementations
```
david-martin: Is that not possible with solution 1?

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+- No need for specific plugins for React and Cordova (platform specific wrappers can be provided) with single code base
```
david-martin: What is the difference between a plugin and a wrapper?
Is there a native element to both? (Or the possibility of a native element, depending on the feature?)

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+- No need to adjust native code for the plugins
```
david-martin: Is this native code the other SDKs (Android & iOS),
or is it plugins that may/may not exist yet?

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+- No challenges related with cross js/native result access.
```
david-martin: I don't understand this. What are the challenges that are avoided/dismissed?

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+
+See investigation details bellow for more details
```
david-martin: 'bellow' > 'below'

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Unanswered questions
```
david-martin: can remove this whole section and add it to the PR description (to avoid accidental merge with it still there)

sdks/Cordova.md
david-martin: Is networking or logging part of the Core SDK on any platform so far?
If not, I think this can be left out of any investigation/proof of concept work ongoing

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+cordova.auth.login()
+// Allow to retrieve token in javascript.
+cordova.auth.getToken()
+```
+
+This code relects how cordova based solutions work in general.
+
+## Investigation process
+
+All investigations and different variations of architecture will be available in following repository:
```
david-martin: Lets rephrase this in a way that will make sense for anyone reading 6 months from now.
e.g.
`Investigations and different variations of architecture were carried out in following repository`

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+
+### POC for Core
+
+Two options were investigated
+1. Wrapping core into plugin. 
+1. Writing pure JS helper.
+
+Wrapping core into plugin was proven difficult and innefective:
+- JavaScript implementation is trivial
+- Droping configuration into native projects may be bad user experience for cordova developers.
```
david-martin: What do you mean by this? Can you give an example of what the developer would have to do, and why its a bad user experience

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+- Core android IOS providing helpers that are available on cordova/javascript
+- Only parser needed to be linked back to javascript
+- The way Android instantiates service will not be compatibile with Cordova
```
david-martin: Got an example?

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+### POC for Core
+
+Two options were investigated
+1. Wrapping core into plugin. 
+1. Writing pure JS helper.
+
+Wrapping core into plugin was proven difficult and innefective:
+- JavaScript implementation is trivial
+- Droping configuration into native projects may be bad user experience for cordova developers.
+- Core android IOS providing helpers that are available on cordova/javascript
```
david-martin: I don't get this point in the context of `Wrapping core into plugin was proven difficult and innefective`

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+- Only parser needed to be linked back to javascript
+- The way Android instantiates service will not be compatibile with Cordova 
+- Wrapping core into cordova plugin will put some limitations on native implementation we want to avoid.
```
david-martin: What do you mean? Got an example?

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+
+JavaScript configuration parser/processor was implemented.
+This was proven to be the most efficient way to resolve configuration parsing problem.
+
+### Interactions between services (binding)
+
+TODO Link github example
+
+### Investigating individual services
```
david-martin: I don't think this is absolutely necessary at this time.
If we keep it more generic and prove & decide on a solution based on:
* getting the Core SDK working
* plus maybe something that touches on a requirement for native code

That should be sufficient for the proposal.

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+### Metrics
+TODO Document
+
+Suggestion/Spike: metrics done in pure JavaScript/Typescript.
+Challenge: How to push from cordova plugins. Can we pull for cordova?
+
+# Possible limitations
+
+Solution will be able to reuse existing cordova plugins like push, 
+however to effective manage and build cordova plugins,
```
david-martin: 'effective' > 'effectively'

sdks/Cordova.md
```diff
@@ -0,0 +1,136 @@
...
+Challenge: How to push from cordova plugins. Can we pull for cordova?
+
+# Possible limitations
+
+Solution will be able to reuse existing cordova plugins like push, 
+however to effective manage and build cordova plugins, 
+both IOS and Android AeroGear Services SDK should not contain any SDK specific code.
+Instead it should just link to the library that can be then also wrapped to the Cordova plugin on different repository.
+
+I identified a couple solutions to extract existing code from SDK,
```
david-martin: As per comment above, lets not get bogged down in the various service SDKs or how existing SDKs (like push) can be used.

david-martin (dismissed):
@wtrocki Reviewed. Apologies for the large number of comments. I'm ramping up and refreshing my memory on some of this.

This is probably the most important comment https://github.com/aerogear/proposals/pull/25#discussion_r169277221

sdks/Cordova.md
wtrocki: Going to remove proposion nr1 as it's kinda confusing. Proposition nr1 has native code wrappers as plugins. Proposition nr2 is js implementation, but it still allows to use plugins when needed so this is where confusion may come from.

See: https://github.com/aerogear/proposals/pull/25#issuecomment-366304423

sdks/Cordova.md
wtrocki: There is only push plugin available. We need to wrap native code into plugins which is not as trivial as use existing JavaScript libraries.

sdks/Cordova.md
wtrocki: `plugin` - native implementation with interface to talk with JavaScript code.
`wrapper` - it's just javascript code that can be provided by users to change some behavior - like custom storage.

sdks/Cordova.md
wtrocki: Going to clarify that. It's native code that it's not the plugin

sdks/Cordova.md
wtrocki: If code in native class changes field in memory then you need to have way to query this field in cordova or rewrite native implementation to not use that.

wtrocki:
@david-martin Great review! Updated content and moved to implementation.

wtrocki:
I have added updates to proposal. As we already have working POC I think it's best to merge that - and keep it minimal for the moment so it will not outdate. 
This proposal was really good to kickstart general conversation and since we already got agreements and collaborated effectively inside the repository I think it's time to seal that effort. Additional work we are doing now (checking native uses cases) can be done in separate proposals)

FYI @david-martin @pb82 @paolobueno

wtrocki:
@david-martin Great refactor for proposal! I really like it.

david-martin:
@wei-lee @pb82 Can you give final ack/feedback before merge?
FYI, repo where proposal investigations happened is https://github.com/aerogear/aerogear-js-sdk

david-martin:
@pb82 I think any actions/updates out of https://issues.jboss.org/browse/AEROGEAR-2064 can modify the proposal after

pb82:
👀

josemigallas:
@wtrocki What does "pure js" mean exactly?

wtrocki:
@josemigallas - Good point! It's actually TypeScript now. Overlooked that one.

josemigallas:
So if have understand it correctly, we will implement "base" JS modules (for example `aerogear-js-core`) that ideally will be imported and used by a cordova/RN/browser application.
Then, in case there are platform-specific, say you need to do some specific setup in ReactNative, we would have specific module, that wraps up the base one (following the example a RN app will import `areogear-js-core-react-native`, that internally uses `aerogear-js-core`).

Am I speculating too much? 😅

wtrocki:
@josemigallas Module will be the same. It will just have specific core or extension points for cordova or react if needed.

josemigallas:
I can't think of any use case for this but my impression is they should be separated (Single Responsibility). If tomorrow we removed support for Cordova or added support for another JS platform, we wouldn't have to touch the "base" module, only remove or add the platform-specific ones. Does this make sense?

pb82:
@wtrocki The proposal doesn't mention Webpack, should it?

wei-lee:
I think the core module it's pretty good and easy to understand. Still not sure about how to extend to support cordova/RN native code. Is it possible to just have a mock service that will demonstrate what the module will be like in order to use a cordova/RN plugin?

pb82:
@wei-lee @david-martin already added an example for react-native, see: https://github.com/aerogear/aerogear-js-sdk/blob/master/example/react-native/src/components/StartScreen.js#L8

josemigallas:
Just in case I didn't explain myself correctly. If we look at the react-native example @pb82 referenced, it imports `@aerogearcatalog/core` module directly. If core would, at any point, make use of any React Native feature or function, we shouldn't implement that there. There should be a wrapper module say core-react-native that uses core internally. Then in the example you would import `@aerogearcatalog/core-react-native` instead.

pb82:
@josemigallas Do you mean that instead of using the output in `dist` we should create a separate package for each reacht native module that wraps this output? It might be required for some modules that can only be implemented natively but the general idea is that the created bundle works for Cordova and react native.

wtrocki:
>  Is it possible to just have a mock service that will demonstrate what the module will be like in order to use a cordova/RN plugin?

Great idea! Working on that already! @wei-lee Do you think we could use 
keycloak.js provided by keycloak to check if will work with react native?
We know that it will work with core as it will be only configuration needed 
and it has already Cordova support.

> The proposal doesn't mention Webpack, should it?

IMHO What we use to wrap things it's not as much important and still up to swap.
We have been evaluating browserify and webpack. Browserify required multiple external tools like tsify etc. which made entire solution more complex. We went with webpack for the moment, but we are still ok to swap to whatever will work in the future. Let's keep proposal separate from this as by the end of the day bundle is delivered and supported. Proposal really opened conversation - low level technical details should be always done in repository.

> Then in the example you would import @aerogearcatalog/core-react-native instead.

That could be option, but without actual use cases it's just to early to talk about it. Push, Sync etc. will require network wrappers for react. Keycloak is being evaluated etc. @josemigallas - It will be cool if you will try out our Cordova example - any suggestions are welcome especially when they are done as PR :)

> Do you mean that instead of using the output in dist we should create a separate package for each reacht native module that wraps this output? 

That's up to future discussion and actual trivial automation/distribution task rather than challenge.  We could first focus on challenges which are at the moment state of the js solutions we are wrapping and possibility to swap implementations to native. Actual delivery, packaging can be adjusted, build when we will know the scope of the changes we going to do, how we integrate them together etc.

josemigallas:
@pb82 What I mean is to not to reference Cordova or ReactNative or whatever frameworks we want to support inside the JS sdk modules.

wei-lee:
> Great idea! Working on that already! @wei-lee Do you think we could use
keycloak.js provided by keycloak to check if will work with react native?
We know that it will work with core as it will be only configuration needed
and it has already Cordova support.

I don't know if we need to use keycloak.js TBH. What I mean is some pseudo code to show what it will be like to invoke a native cordova plugin or RN plugin. it doesn't need to do anything, but it should help to  understand.

david-martin:
> What I mean is some pseudo code to show what it will be like to invoke a native cordova plugin or RN plugin. it doesn't need to do anything, but it should help to understand.

@wei-lee I think what you are asking for is the outcome of this jira, with an example of how it might be solved https://issues.jboss.org/browse/AEROGEAR-2064 :)
I agree if that's the case. It would be good to see that before merging

wtrocki:
@josemigallas Some early version of your concept: 
https://github.com/aerogear/aerogear-js-sdk/pull/16

wtrocki:
@david-martin Are we ok to get that merged? Reviewed all comments. It feels like all concerns were addressed and more detailed/technical work was actually done on the list.

david-martin:
Agreed.
Thread for reference https://groups.google.com/d/msg/aerogear/-C1eMiOLnfQ/bNtkJAh3BAAJ

PR #30, opened 2018-02-28
clarifications about native modules

pb82:
Clarify the usage of native moduels for Cordova and React Native.

### DIFF
--- a/sdks/js-sdk.md
+++ b/sdks/js-sdk.md
@@ -6,7 +6,7 @@ This document outlines what is needed to provide a set of the SDK's that can be 
 
 ### Primary Goals
 
-* Provide SDKs (Core, Metrics, Auth etc..) for Mobile Developers to allow them to integrat Cordova, React Native & Web applications into Mobile Services.
+* Provide SDKs (Core, Metrics, Auth etc..) for Mobile Developers to allow them to integrate Cordova, React Native & Web applications into Mobile Services.
 
 ### Secondary Goals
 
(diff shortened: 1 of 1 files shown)


sdks/js-sdk.md
```diff
@@ -19,14 +19,15 @@ The 3 target platforms (Cordova, React Native & Web) all support JavaScript as t
 
 ## Proposed solution
 
-Provide JavaScript implementations (via a JavaScript SDK). This means the deault way of implementing an SDK inteface will be in Javascript. However, this may not always be possible. Each platform will provide wrappers for specific APIs if needed. Doing this will make Cordova, Web and React Native SDK's more consistent. Implementation could detect platform and load specific React Native/Cordova/Browser functions written in JavaScript or call native plugins if necessary.
+Provide JavaScript implementations (via a JavaScript SDK). This means the default way of implementing an SDK interface will be in JavaScript. However, this may not always be possible. 
+For the Cordova and React Native platforms we will provide a plugin that encapsulates the native code. This plugin conforms to an interface defined in the JS SDK that abstracts all the native platform specific code. Doing this will make Cordova, Web and React Native SDK's more consistent. We will bundle an
```
wtrocki, line 23: [Question] Just to clarify - could we use existing plugins if there are available? Content is right - I just read that as we will always wrap native code into plugins, but we could also reuse.

wtrocki:
LGTM from my point of view - aligns with discussion on the list.

wtrocki (approved):

sdks/js-sdk.md
pb82: An example for this would be push: https://github.com/aerogear/aerogear-cordova-push

I think we should make use of that, explain it in the Push docs and maybe add a thin API in the JS SDK that calls the functions provided there. That sounds more reasonable than copying the code into our native-extensions plugin.

wei-lee (approved):

sdks/js-sdk.md
wtrocki: @pb82 - To correct: "existing third party plugins" - like Cordova Device plugin etc.?

pb82: Good question, but for small things we should probably avoid it. Otherwise there would be to many manual installations for the users.

wtrocki: This plugins may be already available by default in Cordova (there is set of default plugins for cordova that covers 90% of the generic use cases when native code is needed) Bundling this plugins may be more effort. I would say we could relay on proven set of plugins that are provided by cordova/react (if there are those for react) and wrap other cases into our own.

pb82: I remember reading about Cordova moving core functionality into plugins but can't find it now. We will have to deal with it when (if) the JS SDK at some point relies on functionality that is provided in a 3rd party cordova plugin.

wtrocki: This conversation was just for my own understanding. Thanks. Changes look great

PR #31, opened 2018-03-06
✨ UPS Integration with OpenShift Proposal

david-martin:


push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,100 @@
+# Unified Push Server Integration with OpenShift 
+
+## Introduction
+
+This proposal outlines some developer experience issues with provisioning, setting up and configuring the Unified Push Server (UPS) on OpenShift, and integrating it with Apps using the Aerogear SDK.
+
+## Problem Description
+
+The Unified Push Server is made up of a Java application & a database (MySQL).
+It also has a dependecy on an auth provider (currently Keycloak) for users of the UPS Administration UI or REST API.
```
matzew: The REST API is basically two parts
* Management APIs (e.g. everything thats under `/application` or `/metrics` (which really is analytics), that also can be controlled via the UI and its wizard for setting up `PushApplications` and their `Variants`
* _public_ APIs (endpoints for the actual mobile devices (E.g. for [registration](https://aerogear.org/docs/specs/aerogear-unifiedpush-rest/index.html#246535932) or any backend-service for [sending](https://aerogear.org/docs/specs/aerogear-unifiedpush-rest/index.html#397083935))

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,100 @@
...
+There are docker images available for UPS.
+There is also a proof of concept UPS Ansible Playbook Bundle (APB) available.
+However, the APB just provisions UPS and MySQL currently.
+It expects an existing Keycloak instance to be running somewhere with the appropriate Realm, Client & Roles already setup.
+
+The second problem is the lack of a consistent login experience when compared to other services (e.g. Jenkins, Prometheus, Grafana).
```
matzew: these services are all "just" protected using Openshift - not Keycloak, right ?

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,100 @@
...
+As detailed in [Sign On Proposal](../auth/developer-single-sign-on-across-mobile-services.md), the goal is to have a consistent login experience by logging in using OpenShift credentials.
+
+The third problem is the switch in context and terms between creating Mobile Clients & Services in OpenShift, and setting up, configuring & integrating Apps with UPS Push Applications/Variants for Push messages.
+The developer creates a 'Mobile Client' in OpenShift as a logical construct to represent the App they are developing outside of OpenShift.
+In UPS, the developer creates a 'Push Application' and 1 or more 'Variants' depending on the number & type of Apps they are developing outside of Openshift.
```
matzew: not really sure this is a _problem_ I think, it's similar to a user setting up a build pipeline, in Jenkins or a dashboard in Grafana

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,100 @@
...
+These problems don't block the developer, but do make it harder than it could be to add Push integration to their Apps. 
+
+## Terms
+
+- TODO
+
+## OAuth provider support 
+
+To address the first & second problem, UPS will be modified to support different auth providers.
+The 2 providers will be Keycloak (to maintain backwards compatibility for existing users) and a generic OAuth Provider.
```
matzew: > to maintain backwards compatibility for existing users

not just that - also to give the upstream community a chance to continue as normal. Not everybody is ultimately only interested in running the UPS in Openshift. It was designed to run anywhere - not just Openshift.  A tight Openshift integration is the scope of this proposal, as an alternative option to "run it standalone"

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,100 @@
...
+### Syncing
+
+Syncing will happen *from* OpenShift *to* UPS for the following resources:
+
+* MobileClient => Variant
+* Secret (p12 or secret key) => Variant configuration
+
+The Secrets will have an appropriate label so it can be easily filtered when looking for resources to sync.
+The sidecar will overwrite the Variant fields and configuration to match what the values are in OpenShift.
+This means that any changes to a Variant directly in the UPS UI will be overwritten on next sync.
```
matzew: this will happen _only_ on the tight Openshift integration

matzew:
so far, so good

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,100 @@
...
+This proposal outlines some developer experience issues with provisioning, setting up and configuring the Unified Push Server (UPS) on OpenShift, and integrating it with Apps using the Aerogear SDK.
+
+## Problem Description
+
+The Unified Push Server is made up of a Java application & a database (MySQL).
+It also has a dependecy on an auth provider (currently Keycloak) for users of the UPS Administration UI or REST API.
+There are docker images available for UPS.
+There is also a proof of concept UPS Ansible Playbook Bundle (APB) available.
+However, the APB just provisions UPS and MySQL currently.
+It expects an existing Keycloak instance to be running somewhere with the appropriate Realm, Client & Roles already setup.
```
matzew: not really sure I fully understand what exactly the "first" problem is :-)

matzew:
One question, @david-martin: We can not use Keycloak's "Identity Provider" feature (for Openshift), due to missing roles and permission mappings?

push/ups_integration_with_openshift.md
david-martin: Good to know.
I'll update with more specific info

push/ups_integration_with_openshift.md
david-martin: Correct

push/ups_integration_with_openshift.md
david-martin: Good point.
After chatting with @maleck13, I think coming at this from 'Use Cases' rather than 'Problems' might be better.
I intend to rephrase the 'Problem Description' section, but keep the underlying points

push/ups_integration_with_openshift.md
david-martin: Great point.
I'll add more info about this

push/ups_integration_with_openshift.md
david-martin: Correct. You'll only have the sidecar if you're running on OpenShift, so this will only happen then

push/ups_integration_with_openshift.md
david-martin: See comment above where I mention I'll rephrase this section https://github.com/aerogear/proposals/pull/31#discussion_r172800370
What I'm trying to highlight here is the APB currently expects a lot of things to already exist i.e. Keycloak with specific relam & roles.
What we'll do is remove the need for this setup.
A User story that covers this (which I'll add to the doc) might be
'As a developer, I want UPS running in my OpenShift namespace, ready to go without any additional setup so I can start to use it immediately'

david-martin:
> One question, @david-martin: We can not use Keycloak's "Identity Provider" feature (for Openshift), due to missing roles and permission mappings?

@matzew I wouldn't say 'can not', but it would be tricky to do it, plus the experience might be odd.
You'd be logging into UPS via Keycloak via OpenShift. I'm not sure how that would look from the user's point of view.
Plus we'd need to figure out how OpenShift User roles get mapped to Keycloak roles, and where those mapping happens.

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,122 @@
...
+*As a developer, I want to write custom logic for sending Push messages so that I can dynamically send messages based on varying factors and inputs*
+
+Push messages can be sent by calling the [sender API](https://aerogear.org/docs/specs/aerogear-unifiedpush-rest/index.html#397083935) with the correct key. There are features available in the OpenShift and the Catalog that can simplify the provisoining of a custom service, and obtaining secret credentials (like a key) from a provisioned service.
+
+## OAuth provider support 
+
+To address the first & second problem, UPS will be modified to support different auth providers.
+The 2 providers will be Keycloak (to maintain backwards compatibility for existing users) and a generic OAuth Provider.
+The OAuth flow will follow the same pattern & flow used by the [OpenShift OAuth Proxy](https://github.com/openshift/oauth-proxy).
+The UPS APB will need to be modified to include an OAuth ServiceAccount, similar to how [Jenkins uses a ServiceAccount](https://github.com/openshift/jenkins-openshift-login-plugin#browser-access).
```
maleck13: creating service accounts can cause permissions problems and may need elevated perms

push/ups_integration_with_openshift.md
maleck13: I think we need to be careful here, as this means a developer will have to interact with OpenShift object in order to change things  in UPS. This might not be the developer experience we want. 
Perhaps a better approach may be something like:
- if the variant does not exist then it is created in UPS and reflected in the CRD with an the variant ID
- if it does exist the CRD is updated with any details present in UPS
- if there is and ID present in the CRD that is no longer present in UPS then it has been deleted in UPS and should be removed from the CRD

This means developers are free to make changes in UPS where it makes sense to them and the rules of syncing the data are clear. As mentioned probably want a follow on proposal around this.
But as a general principal
- if the resources never existed then create it
- if it exists then update the openshift representation
- if did exist but now doesn't update the openshift representation

maleck13 (commented):
some questions around the synching of data

push/ups_integration_with_openshift.md
david-martin: I don't think we'll have an issue here if we follow what's done for Jenkins.
https://github.com/openshift/origin/blob/master/examples/jenkins/jenkins-persistent-template.json#L186-L195

push/ups_integration_with_openshift.md
david-martin: @maleck13 Here's info on how the syncing works for jenkins (which may influence our thinking) https://github.com/openshift/jenkins-sync-plugin#openshift-jenkins-sync-plugin
It doesn't look like it's very destructive, but I'd like to try it out and poke around more (chaos monkey)

The only think that get's synced back to OpenShift in that case is Job Runs as Build Objects

> changes in a Jenkins Build Run thats associated with a Jenkins Job gets replicated to an OpenShift Build object (which is created if necessary if the build was triggered via Jenkins)

david-martin: Jira task that will look to understand and define this section better https://issues.jboss.org/browse/AEROGEAR-2326

wtrocki:
@david-martin Quick question. Is this something that we should consider for current push SDKs or this will be done in the future?

david-martin:
> @david-martin Quick question. Is this something that we should consider for current push SDKs or this will be done in the future?

It will need to be considered as part for delivering UPS on OpenShift, where the Aerogear SDK includes UPS support.
So it's very relevant to any work with moving the ups sdks into the main sdk repos

push/ups_integration_with_openshift.md
wtrocki: Users will need to pick one of the variants. Sync logic could be complex and can lead to the bugs. I will suggest to let users to specify their variants.

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,122 @@
...
+
+## OpenShift Integration
+
+To address the third problem, a sidecar will be added to the UPS APB.
+This sidecar will provide background bi-directional syncing of resources and metadata between OpenShift and UPS.
+The sidecar will be written in Go and utilise a kubernetes client for watching & updating Kubernetes resources.
+It will use a SerivceAccount for auth when watching resources.
+The sidecar will also use a ServiceAccount to access the UPS Admin REST API for reading & updating UPS resources.
+
+### Syncing
```
wtrocki: Users will need to pick one of the variants available in config. Sync logic could be complex and can lead to the bugs. I have limited knowledge on this, but could we use the same aproach as upstream: 

Let users to specify their variants in the code as this is something they need to anyway when more than one is present? 

As for IOS plist > than json for storing secrets.
Upstream is designed to use plist and moving to JSON may be a downgrade.

Changing format and api may also affect migration paths from previous versions or RHMAP and upstream to AeroGear Services (5.0)

@matzew - Do we care to keep the API the same?

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,122 @@
...
+* Variant push configuration (JSON) => MobileClient annotation
+
+In UPS, a JSON object with Push configuration for a Variant is available.
+However, this JSON object isn't part of the generated mobile-services.json file from the Mobile CLI or UI.
+This JSON config will be read from UPS and synced back as an annotation on the associated MobileClient in OpenShift.
+Any manual changes this annotation will be overwritten by the sidecar when it next syncs.
+
+```json
+{
+  // TODO: example mobile-services.json structure with a single variant
```
danielpassos: Syncing with @wtrocki about push, we were thinking if we really need to provide this one. Today the developers need to take care of it yourself, they need to go the UI and copy and paste the JSON provided by the UPS. It also kills the problem with more than one variant.

push/ups_integration_with_openshift.md
wtrocki: TL;DR - just wanted to flag that there is the option to not do any sync, but I'm not sure if that something we want to do. I think it's essential to get json example bellow because from SDK point of view it's not obvious what this will look like.

david-martin: > TL;DR - just wanted to flag that there is the option to not do any sync, but I'm not sure if that something we want to do. I think it's essential to get json example bellow because from SDK point of view it's not obvious what this will look like.

@wtrocki @danielpassos Do you see value in providing UI in Openshift for creating the Variants instead? i.e. the Openshift Web Console has some of the UPS Admin UI parts added to it, and calls the UPS Management API.

push/ups_integration_with_openshift.md
david-martin: @danielpassos @wtrocki Do you see value in any element of syncing?

Do you think the goal should be to have a single config file (mobile-services.json) with config for all services, including UPS?

push/ups_integration_with_openshift.md
danielpassos: I think so. This is how it works today, and It works pretty well. I think to find a way to provide it on the mobile-config.json is not going to be easy.

push/ups_integration_with_openshift.md
danielpassos: I think we can have push included in mobile-services.json, but I don't know if we should provide the variants information there, or if so, how to provide this information there.

I think we can have at least the push URL or so only to _flag_ there is a push services there. I'm not 100% sure about that yet. Something like...

```json
{
    "id": "push",
    "name": "push",
    "url": "http://someurl/push",
    "type": "push",
    "config": {
    }
}
```

If we decide to provide the variants we need to think what is the best way to provide it (talking about apb provide it and SDK to comsume it.

```json
{
    "id": "push",
    "name": "push",
    "url": "http://someurl/push",
    "type": "push",
    "config": {
        "android": [
            {
                "name": "driver",
                "variant": "123456789-123456789",
                "secret": "987654321-987654321",
                "sender": "9999"
            },
            {
                "name": "passeger",
                "variant": "123456789-123456789",
                "secret": "987654321-987654321",
                "sender": "9999"
            }
        ],
        "ios": {
            "name": "admin",
            "variant": "123456789-123456789",
            "secret": "987654321-987654321"
        }
    }
}
```

But how to create/edit/remove the variants informations?

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Problem Description
+
+Rather than highlighting problems, User Stories will be used to capture improvements and value that can be added that makes for a better overall experience when using UPS on OpenShift
+
+*As a developer, I want UPS running in my OpenShift namespace, ready to go without any additional setup so I can start to use it immediately*
+
+The Unified Push Server is made up of a Java application & a database (MySQL).
+It also has a dependecy on an auth provider (currently Keycloak) for users of the UPS Administration UI or REST API.
+There is a proof of concept UPS Ansible Playbook Bundle (APB) available.
```
danielpassos: Should we link APB?

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,122 @@
...
+*As a developer, I want a consistent login experience across all Mobile Services so that I can easily login with the same credentials to administer my services*
+
+As detailed in [Sign On Proposal](../auth/developer-single-sign-on-across-mobile-services.md), a goal with all Mobile Services is to have a consistent login experience by logging in using OpenShift credentials.
+This currently isn't the case as a separate User in Keycloak is used.
+
+*As a developer, I want a simplified setup of Push variants in UPS so that I don't have to manually define them and link them back to Mobile Clients in OpenShift*
+
+Currently there is a switch in context and terms between creating Mobile Clients in OpenShift, and setting up Push Applications & Variants in UPS.
+The developer creates a 'Mobile Client' in OpenShift as a logical construct to represent the App they are developing outside of OpenShift.
+In UPS, the developer creates a 'Push Application' and 1 or more 'Variants' to represent the App they are developing outside of OpenShift.
```
danielpassos: The application can be 1 per project, so it can be created automatically, but what about the variants?

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,122 @@
...
+There is duplication of logical constructs and potential confusion between what Mobile Client is and what a Push Application is.
+
+The developer also has to somehow get the Push configuration from UPS and place it into their App for consumption by the SDK. With the mobile-services.json approach being used by the Mobile CLI/UI and the SDKs for configuration, there's some things that can be done to simplify what the developer has to do here.
+
+*As a developer, I want to write custom logic for sending Push messages so that I can dynamically send messages based on varying factors and inputs*
+
+Push messages can be sent by calling the [sender API](https://aerogear.org/docs/specs/aerogear-unifiedpush-rest/index.html#397083935) with the correct key. There are features available in the OpenShift and the Catalog that can simplify the provisoining of a custom service, and obtaining secret credentials (like a key) from a provisioned service.
```
danielpassos, line 36: How will the developer get the `PushApplicationID:MasterSecret` to send the messages?

push/ups_integration_with_openshift.md
wtrocki: To extend Passos response I kinda feel that services running on openshift should have their own UI, however if we do not want to update or maintain current UI then this may be some option.  I see people just using current push UI the same way grafana is used. Maintaing two separate UI is not convenient. This will also allow us to reuse implementation tests and documentation from upstream.
We did not write our own keycloak dashboards etc.

[Unpopular opinion] Generally putting some end user facing UI into "Platform as service" solution seems to be strange and we should avoid that. What I will expect from OpenShift UI is to see some architecture components like services, mobile apps etc. Push configuration is too big stretch.

push/ups_integration_with_openshift.md
wtrocki: +1 for only url in config. 
I do not see value in sync - it will introduce a lots of additional work for very little value. Users can provision apb and use the same configuration steps as upstream

push/ups_integration_with_openshift.md
david-martin: can do

david-martin:
@wtrocki @danielpassos Some really good points in the comments above.
After chatting with passos offline, I have more understanding of where you both are coming from with the above feedback.

So, there is *no* value in creating a variant without having the associated credentials also (e.g. p12, master secret).
Therefore, I am questioning the value of automatically syncing mobile clients to variants as soon as a mobile client is created.
However, is there value in syncing after they developer has created variant?

For example:
* developer creates a Cordova App (mobile client) in OpenShift
* developer provisions UPS
* developer sets up the fcm/android credentials for push for the Cordova App via *new* UI screens in OpenShift
* this creates an android variant in UPS
* the variant config is retrieved from UPS and stored in/with the Cordova App resource (mobile client)
* the generated mobile-services.json file includes the correct variant config

If the developer comes back later to setup the ios variant for their Cordova app, they can go through a similar process:
* developer sets up the ios credentials for push for the Cordova App via *new* UI screens in OpenShift
* this creates an ios variant in UPS
* the variant config is retrieved from UPS and stored in/with the Cordova App resource (mobile client)
* the generated mobile-services.json file includes the correct variant configs for both android & ios now.

Thoughts?

push/ups_integration_with_openshift.md
david-martin: > But how to create/edit/remove the variants informations?

See latest comment with a suggested flow which shows were/when this could be created.

push/ups_integration_with_openshift.md
david-martin: See suggestion in latest comment https://github.com/aerogear/proposals/pull/31#issuecomment-372456779
I think the variants have to be manually created/linked back to a mobile client.
A mobile client could have multiple variants associated (e.g. cordova).
By linking back to a mobile client, it allows the correct mobile-services.json to be generated

push/ups_integration_with_openshift.md
david-martin: When the developer 'binds' to UPS, a secret will be created that has the master UPS secret value.
This secret can be mounted into a custom service or set as an env var for a custom service.

push/ups_integration_with_openshift.md
david-martin: > [Unpopular opinion] Generally putting some end user facing UI into "Platform as service" solution seems to be strange and we should avoid that. What I will expect from OpenShift UI is to see some architecture components like services, mobile apps etc. Push configuration is too big stretch.

@wtrocki what is your view of the Jenkins integration in Openshift, where BuildConfigs, Builds & Build Pipelines are a UI in front of Jenkins Jobs, Job Runs and Jenkins Pipelines?

wtrocki:
This may be not welcomed opinion, but personally I see that we could manage and create variants in Push admin UI without any sync or UI integration with OpenShift.
I might miss some facts or stakeholders requirements, but this will be pretty simple - provisioned service deals with the all operations using admin ui.

It takes the same time to copy created variant from the admin ui than generate new config. I do not really see any improvement here. It's best to use **plist** for config.

Config is generated per app-id so we will do not need to have android and IOS namespaces.

push/ups_integration_with_openshift.md
danielpassos: Only a note: *not only Cordova* can/should have more than one variant. I have a real Android app using more than one

push/ups_integration_with_openshift.md
matzew: > @matzew - Do we care to keep the API the same?

@wtrocki I'd think that - for iOS - `plist` (like we currently do) is more natrual for _true_ iOS devs

matzew: > @wtrocki what is your view of the Jenkins integration in Openshift, where BuildConfigs, Builds & Build Pipelines are a UI in front of Jenkins Jobs, Job Runs and Jenkins Pipelines?

@david-martin @wtrocki I think there is generaly pros and cons w/ both - our own view could be more `aligned` w/ Openshift that way, just talking HTTP w/ the actual services. which is nice, but also people might be already very familiar w/ the existing UI (which is likely the case on Jenkins, Prometheus and Grafana) - push is a different beast here, since not that known/needed - the others are very generic and popular services

matzew:
> So, there is no value in creating a variant without having the associated credentials also (e.g. p12, master secret). ... 
> Therefore, I am questioning the value of automatically syncing mobile clients to variants as soon as a mobile client is created. 

I think this variant creation should be done only, when push service is weird/bind/conntected to that MAR (Mobile Application Representation). Cordova/Xamarin are different here, as their "unified code" runs on different platforms, so needs very likely APNS and FCM for push - while pure iOS/Android are logically tied to just one variant

matzew:
> However, is there value in syncing after they developer has created variant?

yes - after the MAR has some _relationship_ on push, the sync is pretty useful, and needed - I like the flow you outlined here

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,122 @@
...
+The Unified Push Server is made up of a Java application & a database (MySQL).
+It also has a dependecy on an auth provider (currently Keycloak) for users of the UPS Administration UI or REST API.
+There is a proof of concept UPS Ansible Playbook Bundle (APB) available.
+However, the APB just provisions UPS and MySQL currently.
+It expects an existing Keycloak instance to be running somewhere with the appropriate Realm, Client & Roles already setup.
+This can be improved to eliminate any additional setup or prerequisites required from the developer.
+
+*As a developer, I want a consistent login experience across all Mobile Services so that I can easily login with the same credentials to administer my services*
+
+As detailed in [Sign On Proposal](../auth/developer-single-sign-on-across-mobile-services.md), a goal with all Mobile Services is to have a consistent login experience by logging in using OpenShift credentials.
```
matzew, line 22: should we mention `oauth-proxy` approach ?

push/ups_integration_with_openshift.md
matzew: > The application can be 1 per project, so it can be created automatically, but what about the variants? 

@danielpassos yes, per used namespace -> that's representing exactly one `PushApp` in UPS

push/ups_integration_with_openshift.md
matzew: oauth-proxy, with no-auth UPS behind

push/ups_integration_with_openshift.md
matzew: we still can generate some mobile-services.plist file, no ? and make the iOS SDK aware of that ?

push/ups_integration_with_openshift.md
matzew: I'd rather see that per MAR / backend serivce.

We should NEVER, EVER, expose the MasterSecret, which allows you to send push notifications through the API, on the device config - on the device, only _ITS OWN_ variant and variantSecret should be there

push/ups_integration_with_openshift.md
david-martin: Yes, I'll update based on recent decisions

push/ups_integration_with_openshift.md
david-martin: @matzew Agree completely on never having the master secret on device (not sure where that was suggested).
The master secret could be used by a custom service though for dynamically sending push messages via the sender api

david-martin:
@danielpassos @wtrocki @matzew 
After more feedback and some discussions with @wtrocki, here's some more refined thoughts.

So, I'm still suggesting everything from my previous comment https://github.com/aerogear/proposals/pull/31#issuecomment-372456779
with some modifictions/additions:
* It should be possible for a developer to create multiple Variants from OpenShift UI (regardless of what the Mobile Client is)
  * e.g. development variant for ios, release variant for ios, variant for android
* It should be possible for a developer to 'link' 1 or more Variants with a Mobile Client from the OpenShift UI
  * e.g. A cordova client may need a variant for ios, android and windows
* It should be possible for a developer to change which Variant(s) are linked with a Mobile Client from the OpenShift UI
* The generated mobile-services.json file will contain the config for all variants that are currently linked with the Mobile Client

I think allowing the developer to manually control the link between clients and variants is important for flexibility during development.
We need this link(s) in order to tie back any push activity to a Mobile Client in OpenShift.

One potential point of contention with using mobile-services.json for iOS is that it's different than what the developer might expect i.e. a plist file

EDIT: @wtrocki @danielpassos @matzew  I've pushed a rewrite for the OpenShift Integration section. It's mostly what I'm saying above, but it does include example mobile-services.json files

push/ups_integration_with_openshift.md
david-martin: updated. I added info about the oauth proxy to the auth provider section rather than here in the problems section

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,144 @@
...
+
+To address the third story, a number of UI screens and integration points will be added to the OpenShift UI & the UPS APB.
+
+Here is an example workflow showing the UI touchpoints and the idea of 'linking' Variants to MobileClients in OpenShift:
+
+* developer creates a Cordova App (MobileClient) in OpenShift
+* developer provisions UPS from the Catalog
+* developer sets up the FCM/Android credentials for push for the Cordova App via *new* UI screens in OpenShift
+* this creates an Android variant in UPS (details of how this happens TBD) and links it to the Mobile Client (details of how this 'link' works TBD)
+* the Variant config is retrieved from UPS and stored in/with the Cordova App resource (MobileClient)
```
pb82: This means we have two sources of configuration for variants. See this [comment](https://issues.jboss.org/browse/AEROGEAR-2326?focusedCommentId=13547700&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13547700) for why this is a problem.

david-martin: The 2 main options:

1) OpenShift as the source of truth:
* Syncing *From* OpenShift *to* UPS
  * New Secrets mean a Variant is created
  * Updates to Secrets overwrite the Variant
  * Deleting a Secret will delete the Variant
  * Deleting a Variant will cause it to be recreated (look for variant ID, as stored on the Secret as annotation)
* Syncing *From* UPS *to* OpenShift
  * Variant config will be annotated on the Secret

2) UPS as the source of truth:
* Syncing *From* UPS *to* OpenShift
  * New Variants are *not* synced to OpenShift
  * Updates to Variants that are backed by a Secret get synced back to the Secret
  * Deleting a Variant that is backed by a Secret will also delete that Secret
  * Variant config will be annotated on the Secret
* Syncing *From* OpenShift *to* UPS
  * New Secrets mean a Variant is created
  * Updates to Secrets *don't* get updated in Variants
  * Deleting a Secret has no action in UPS (Variant remains)

Option 1 problems are:
* making changes in the UPS UI will be overwritten on next Sync

Option 2 problems are:
* updating the Secret with new FCM/Apple credentials don't get propagated to UPS



What about a 3rd option where changes are synced both ways?
The conflict resolution strategy could be kept simple where the last in wins.
This could cause confusion (and possibly errors if something is deleted) in edge cases where both the Variant and Secret are modified at the same time.
However, maybe that's OK.

@maleck13 thoughts?

matzew: +1 on Openshift as the source of truth

david-martin: After call with @matzew & @pb82, and comparing to how jenkins manages the syncing between openshift and jenkins pipelines & builds, we decided to go with option 1 for simplicity at this time

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,190 @@
...
+This would be required if running UPS as a shared service.
+That would most likely require changes to UPS to call OpenShift to check user permissions based on the token header, and only showing that developers Push Applications.
+
+These changes will mean UPS will no longer have a dependency on Keycloak when running in OpenShift.
+Also, the developer doesn't have to know about or be concerned with where or how the auth provider is setup and configured for UPS.
+
+## OpenShift UI Integration
+
+To address the third story, a number of UI screens and integration points will be added to the OpenShift UI & the UPS APB.
+
```
david-martin, line 56: @matzew @pb82 I've rewritten most of this section, if you could review.

pb82: 👀

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,190 @@
...
+Push Variant credentials will be stored in OpenShift as a Secret.
+The Secret will have:
+
+* a label to identify it as storing Variant credentials
+* an annotation to link to a specific Mobile Client `aerogear.org/mobile-client-id`
+* an annotation to store the ID of the Variant in UPS `aerogear.org/variant-id`
+* an annotation to store the Variant config from UPS `aerogear.org/variant-config-json`
```
pb82: If we store the variant json itself then we don't need to store the individual properties that make a variant in the secret. We would only submit them once (on bind), get back the variant json from UPS and store that in the secret.

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,190 @@
...
+
+## OpenShift UI Integration
+
+To address the third story, a number of UI screens and integration points will be added to the OpenShift UI & the UPS APB.
+
+Push Variant credentials will be stored in OpenShift as a Secret.
+The Secret will have:
+
+* a label to identify it as storing Variant credentials
+* an annotation to link to a specific Mobile Client `aerogear.org/mobile-client-id`
```
pb82: How would the mobile client discover the secrets associated with it? Should this be a label instead or can we also write selectors over annotations?

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,190 @@
...
+* Syncing *From* UPS *to* OpenShift
+  * Variant ID will be annotated on the Secret after creation
+  * Variant Config will be annotated on the Secret and kept up to date (UPS is the source of truth for this config)
+
+Here is an example workflow showing the UI touchpoints:
+
+* developer creates a Cordova App (MobileClient) in OpenShift
+* developer provisions UPS from the Catalog
+* developer sets up the FCM/Android credentials for push for the Cordova App via *new* UI screens in OpenShift
+* this creates a Secret to store those credentials
```
pb82, line 106: we actually don't need to store the credentials in the secret if we are storing the returned variant definition.

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,190 @@
...
+  ]
+}
+```
+
+A follow on flow would be:
+
+* developer sets up iOS credentials for push for the same Codova App via *new* UI screens in OpenShift
+* these credentials as stored in the Secret
+* on next sync (in the sidecar), an iOS variant is created in UPS
+* the Variant config is retrieved from UPS and stored as annotations in the Secret
```
pb82: Same secret or a secret for every variant?

push/ups_integration_with_openshift.md
```diff
@@ -0,0 +1,190 @@
...
+          "variantId": "some-variant-id",
+          "variantSecret": "some-secret"
+        }
+      }
+    }
+  ]
+}
+```
+
+The developer is free to choose which Variants, from all available Variants in the Push Appplication in UPS, are linked to each Mobile Client.
```
pb82: Isn't the linking done when the annotations/labels are added to the secrets associating them with a Mobile Client? Would this mean that a user can later link an existing configuration (and thus it's secret) to a different mobile client?

push/ups_integration_with_openshift.md
david-martin: different secret for each variant. I'll reword this.
"these credentials are stored in a new Secret"

push/ups_integration_with_openshift.md
david-martin: This section needs revision.
My thinking is we could allow the user to see all variants in the OpenShift UI and let them pick & choose which ones are linked.
I think a first pass would be to allow 1 type of each allowed variant for the Mobile Client type.
To switch between e.g. a development and release certificate, the developer can upload to replace the cert in the variant.

A follow up change could allow switching between known variants (using `oc edit` or new UI screens)
This means there could be variants in UPS that aren't linked to a Mobile Client.
This seems OK to me to allow that, but needs more thought on the execution.

push/ups_integration_with_openshift.md
david-martin: True, however, i think it's best to store them in case we need to re-create/update the Variant.
If the sidecar is watching this resource for changes, it can pick up on a new uploaded cert or changed key, then propagate that change to UPS.
This will mean we can keep the logic for managing variants in a single place (i.e. in the sidecar).
A variant will only every get created by the sidecar.
If we say a 'bind' to the ups apb sets up the variant, then the 'bind' is just creating a Secret with the correct label and data for the sidecar to work with.

push/ups_integration_with_openshift.md
david-martin: Label makes more sense. Updating

push/ups_integration_with_openshift.md
david-martin: See comment here https://github.com/aerogear/proposals/pull/31/files#r175822881

pb82 (approved):

PR #34, opened 2018-03-26
add proposal for app network security

wei-lee:


security/mobile-app-network-security.md
```diff
@@ -0,0 +1,120 @@
...
+1. In AeroGear SDKs, the HTTP libaries (there is only one for each platform) will be configured using the SDK config file (`mobile-services.json`). There will be a new section like this:
+
+    ```
+    {
+      ...,
+      "https": {
+        "selfSignedCerts": {
+          "enabled": true,
+          "keystores": [{
+            "path": "assets/dev.keystore"
```
aidenkeating, line 104: More questions than review stuff (I have a knowledge gap here so excuse any nonsense):

* The `assets/dev.keystore` file is also generated for the user?
* Will this name/path be configurable by the end user?
* I guess the format of this file will be different per platform? What will it look like with something like Cordova? Do you see anything getting more complex here?

security/mobile-app-network-security.md
```diff
@@ -0,0 +1,120 @@
...
+
+In this document we will discuss some of the common problems around network security for mobile developers consuming AeroGear services in their apps, and propose a solution that will help resolve those problems.
+
+We will first look at some of the use cases in order to define the problems, and then look at some of the options to solve them. Finally the proposed solution will be presented.
+
+# Goals
+
+## Primary Goal
+
+* Reach agreement on a solution that can help mobile developers properly handle network security in their mobile apps that consume AeroGear services
```
aidenkeating: Just to confirm. This is only for network security between the app and AeroGear services. Not between an app and any other stuff it communicates with?

security/mobile-app-network-security.md
wei-lee: good questions.

> The assets/dev.keystore file is also generated for the user?

yes, it will be. But only for certain platforms (mainly android).

> Will this name/path be configurable by the end user?

I would think so.

> I guess the format of this file will be different per platform? What will it look like with something like Cordova? Do you see anything getting more complex here?

I am only thinking about ios and android here. I don't think we will be able to support self-signed certs for Cordova at all. Mainly because we will not be able to modify the configurations for the HTTP client in the webview. 

 I will also update the proposal to answer those questions.

aidenkeating: Ah yep that makes sense on the Cordova one, thanks

security/mobile-app-network-security.md
wei-lee: yes, that's right.

security/mobile-app-network-security.md
```diff
@@ -0,0 +1,126 @@
...
```
david-martin: You can probably remove 'Reach agreement on' as this PR review is what gets the agreement. 
Just have 'A solution that ....'

security/mobile-app-network-security.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Secondary Goal
+
+* Provide use cases and try to explain the problems to audiences that may not be familiar with the area, to help them understand and maybe decide the business value of the solution
```
david-martin: This sounds like a goal of the PR rather than a goal of doing what the proposal is proposing.
Does that make sense?

Unless I'm misunderstanding.
Who is the 'audiences' in this goal?

security/mobile-app-network-security.md
```diff
@@ -0,0 +1,126 @@
...
+* Self-signed Certificate: A self-signed certificate is an identity certificate that is signed by the same entity whose identity it certifies.
+* Man-in-the-middle attack: It is an attack where the attacker secretly relays and possibly alters the communication between two parties who believe they are directly communicating with each other.
+
+# Use Cases
+
+The main use cases that will be discussed in this document are:
+
+1. As a mobile developer, I want to use AeroGear services in my mobile apps. I have the AeroGear service running locally with HTTPS, and I want to test against the local services from my mobile app on an actual device.
+
+2. As an enterprise mobile developer, I need to develop a mobile app for internal use. I need to make sure my mobile app can connect to AeroGear services that are running with the organisation's own HTTPS certificates.
```
david-martin: Should we use the terms from above and say 'organisation's Self-signed certificate' instead of 'organisation's own HTTPS certificates'

security/mobile-app-network-security.md
```diff
@@ -0,0 +1,126 @@
...
+
+Next, he set up a proxy server using tools like [Charles proxy server](https://www.charlesproxy.com/). He install the proxy server's certificate on his own device, and pointed the device to use the proxy server. Since the proxy server's certificate is now trusted on his device, the HTTP client will allow HTTPS connection to the proxy server, and user A can see all the traffic data between the client and the server. With this information, user A is now able to initiate attacks against the bank's backend system.
+
+In this example, the user of the app wants to become the attacker. But more seriously, this type of attacks can be performed without permissions from users. For example, a hacker can setup a malicious WIFI hotspot, acting like a proxy server, and ask users to install the self-signed certficates to their devices in exchange for the free connection. If users don't know what they are doing, they could endup installing the self-signed certificates and connecting to the WIFT, and thus give the hacker the opportunity to eavesdrop all the HTTPS traffic, and steal important data, without them realising it.
```
david-martin: 'WIFT' > 'WIFI'

security/mobile-app-network-security.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Problems Explained
+
+If you already know quite well about the problems, then you can skip this section. 
+
+### Problem 1: Support self-signed certificates
+
+To consume AeroGear services, mobile apps will use the AeroGear SDKs, which use HTTP/HTTPS to communicate with the backend services. In most cases, HTTPS is used to ensure the security of the data during transimission. 
+
+When the HTTPS connection is established, the HTTP client on the device will need some way to verify the the authencitiy of the server. This is done by checking the server's certificate during the handshake between the client and the server.
```
david-martin: 'HTTP client' > 'HTTPS client'

security/mobile-app-network-security.md
```diff
@@ -0,0 +1,126 @@
...
+
+In most cases, the server should have a certificate that is signed by a trusted CA to confirm the identity of the server. The HTTP client can then use the trusted CAs' certificates that are pre-installed on the device to verify the server's certificate.
+
+However, due to various reasons (like cost, and sometimes it is just not necessary), CA-signed certificates are not always being used for HTTPS connections. Instead, self-signed certificates are being used. But in this case, the HTTP client on the mobile device will not be able to verify the authentity of the server, and it will refuse to connect to the server. 
+
+In use case #1, when the AeroGear services are running locally, they will be using self-signed certificates by default. This means developers will not be able to connect to these services from their mobile devices by default.
+
+To solve this problem, all the modern OSes allow users install self-signed CA certificates on their devices and trust them. However, there is a side effect: it is possible for attackers to take advantages of it and perform man-in-the-middle attacks.
```
david-martin: 'all the modern OSes' > 'Android & ioS' ?

david-martin (approved):
Approved with some minor comments/suggestions. The main problems and solutions are detailed well

security/mobile-app-network-security.md
wei-lee: Not sure I get it. Is there a problem with this secondary goal, or do you think this proposal doesn't meet this goal?

The `audiences` here are anyone who is not technical or familiar with the area.

security/mobile-app-network-security.md
wei-lee: sure

security/mobile-app-network-security.md
wei-lee: should be `all the modern mobile OSes` and yes there are more than `Android and iOS`

security/mobile-app-network-security.md
david-martin: Does reaching this goal mean that we've produced content (docs, videos, examples etc..) that try to explain the problems for people who aren't familiar with them?

security/mobile-app-network-security.md
david-martin: But do we intend to, as part of doing this proposal, solve the problem for anything other than Android and iOS?

security/mobile-app-network-security.md
wei-lee: right, no. Both goals are only for this proposal. That's why there is a `problem explained` section in this proposal to try explain the problems in a bit more detail.

security/mobile-app-network-security.md
wei-lee: no. But I don't see it is a problem, as this statement isn't really about our solution? 

I don't really mind changing to `Android & iOS` if you have a strong opinion about this.

security/mobile-app-network-security.md
david-martin: Understood.
So I think usually the Goals of proposals are the outcome of doing what the proposal is suggesting, if you get what I mean.

e.g. 'Allow Mobile Developers using the Aerogear SDK to pin certificates for Aerogear Services being used'

with a secondary goal of maybe 'Allow Mobile Developers using the Aerogear SDK to pin certificates for custom services/runtimes'

I'm not saying these are the goals, just that the outcome of doing this proposal isn't completely clear to me

wei-lee: Ok, that is not how I understand what are the "goals". For now, I will change the title to "Goals of the proposal" to make it clear

security/mobile-app-network-security.md
david-martin: no strong opinion given it's not about our solution :+1:

security/mobile-app-network-security.md
david-martin: perfect, thanks

aidenkeating (approved):

wei-lee:
no further feedback, going to merge this for now.

PR #35, opened 2018-04-24
Update for the JS SDK docs

paolobueno:
# Motivation

https://issues.jboss.org/browse/AGCORDOVA-215

Document the current state of the JS SDK, update filesystem structure and add architecture diagram.


sdks/js-sdk/README.md
```diff
@@ -0,0 +1,150 @@
+# Abstract
+
+This document outlines what is needed to provide a set of the SDK's that can be used for JavaScript platforms, specifically Cordova, React Native & Web.
+
+## Goals
+
+### Primary Goals
+
+* Provide SDKs (Core, Metrics, Auth etc..) for Mobile Developers to allow them to integrate Cordova, React Native & Web applications into Mobile Services.
```
josemigallas: I want this to be true, however.. should we move React Native to the secondary goals, along with other platforms like flutter?

sdks/js-sdk/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+The 3 target platforms (Cordova, React Native & Web) all support JavaScript as the development language. To develop Mobile Apps that integrate into Mobile Services, it is necessary to provide SDKs for Mobile Developers for the target platform & language. These SDKs need to be written & maintained. 
+
+## Proposed solution
+
+Provide JavaScript implementations (via a JavaScript SDK). This means the default way of implementing an SDK interface will be in JavaScript. However, this may not always be possible. 
+For the Cordova and React Native platforms we will provide a plugin that encapsulates the native code. This plugin conforms to an interface defined in the JS SDK that abstracts all the native platform specific code. Doing this will make Cordova, Web and React Native SDK's more consistent. We will bundle an
+implementation of that Interface for the Web Platform in the JS SDK. The plugins for Cordova and React Native will be separate repositories. The JS SDK requires the users to install the plugin for those Platforms and it will either try to auto-detect the platform, or require the user to pass in a `PlatformImpls` module when a service is instantiated.
```
josemigallas: Should we be more categorical here and say the SDK **will be able to detect** the platform?

sdks/js-sdk/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Advantages
+
+- Ability to reuse code for Cordova, React Native & Web and work in a single repository
```
josemigallas: I would add _... and any other JS-based framwork._

sdks/js-sdk/README.md
```diff
@@ -0,0 +1,150 @@
...
+- Can reuse JavaScript implementations for keycloak, push, sync.js etc.
```
josemigallas, line 30: Does this mean to have a single implementation in JavaScript for each service regardless of platform?

sdks/js-sdk/README.md
```diff
@@ -0,0 +1,150 @@
...
+- No need to adjust native code (from Native SDKs) for plugin use. The native wrappers will not depend on the Android or iOS SDKs but provide their own implementations.
+
+## Architecture
```
josemigallas, line 33: It could be nice to have a header per package (core, other sdks...) and explain that core would be required by the rest of packages.

sdks/js-sdk/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+Despite the built-in `JSON` API to parse the configuration file, the Core SDK will contain a dedicated module to parse the configuration file, this has the following benefits:
+
+- Better documentation of the configuration values via TypeScript interfaces
+- Encapsulation of changes in the config file structure
+- Possible support for dynamic and asynchronous fetching of future configuration values
+
+### Metrics
+
+As with the other SDKs, Core will contain the implementation for metrics gathering and reporting, providing a JS interface for uploading records which must also be used even by modules that rely heavily on native implementations, such as some of the Self-Defence Security Checks.
```
josemigallas, line 45: Not sure about this, @wtrocki and I thought we could have a separated module for metrics, and core would only contain the common interfaces, abstract classes and config helper.

sdks/js-sdk/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+The uploading of metrics records utilizes the networking layer provided by Core.
+
+### Individual SDKs
+
+Depending on the SDK, a TypeScript/JavaScript implementation may be possible. In cases where it isn't, existing plugins can be used, or new plugins can be created for services.
+
+## Adding a module to the JS SDK
+
+New modules written JavaScript or Typescript will have the following items:
```
josemigallas, line 55: Should we be more categorical here and say only Typescript?

sdks/js-sdk/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+- A single `index` entry file that organizes and exposes all the exported contents of the module.
+- Manually written or compiled files for a Cordova plugin should be present in the `www/` folder by convention
```
josemigallas, line 58: ~~I think this is outdated and contradictory with the next paragraph. It's saying here where to put Cordova code inside a module but below it says the plugin is should be in a separated plugin, IMO the correct one is the latter. Or am I missing something?~~

Edit: after reading the whole "Adding a module to the JS SDK"

I think the proposal is very good and clear until this point. Not that this section is bad, but I think maybe it is not what the proposal is to address (plus it is a bit outdated). I mean, Cordova plugins and React Native modules have a standard way of being developed and it's described in their respective docs. I think we should simply say that native interfaces has to be implemented separately from the SDK module, and that the module needs to have an index, tests, a src folder, etc...

wtrocki:
@paolobueno As far I remember proposal matches what we have at the moment. We could add diagram I have linked in your ticket. WDYT?

sdks/js-sdk/README.md
```diff
@@ -0,0 +1,150 @@
...
+Native bindings for React Native should also be delivered as a separate NPM package and register their native modules in react-native's `NativeModules` object.
+The code in the 'main' JavaScript module should be able to check for the module's presence in order to react accordingly.
+
+Optionally these can include typing declarations or convenience JavaScript to facilitate access, but these should still allow for feature-checking by the main module.
+
+## Investigation
+
+A Proof of Concept was done in the following repository: https://github.com/aerogear/aerogear-js-sdk
+
+### PoC results for the Core Module for the JS SDK
```
wtrocki: This kinda shows the thinking progress. We can revert that if needed by let's remove it!

sdks/js-sdk/README.md
paolobueno: I'd say Web captures that. With WebAssembly and other compile-to-js solutions it's actually more than just js-based frameworks :wink:

sdks/js-sdk/README.md
paolobueno: The other sdks have metrics as part of core so that other modules can report metrics without extra explicit dependencies, i.e. metrics is part of the 'common interfaces and classes' as much as centralized logging is.

From what I see in the js-sdk repo this is the case right now

sdks/js-sdk/README.md
paolobueno: Moving RN to secondary. Flutter is dart-based but it definitely looks similar to RN, I'd love to see this sdk being 'upgraded' to a generic hybrid one instead of just JS, but best not to add this to the proposal yet.

sdks/js-sdk/README.md
josemigallas: Fair enough!

paolobueno:
@wtrocki I've updated the diagram to add a custom plugin instead of only already-available ones: https://github.com/aerogear/proposals/blob/69117890c1b9d16fea3711ce8f34bae314ff7d93/sdks/js-sdk/img/diagram.svg.

The previous version of the diagram is available in c7a1713a1f203c56255604c4839f9758a6130863 if we decide to revert

sdks/js-sdk/README.md
wtrocki: yeah. Not including parts that call native bits.

sdks/js-sdk/README.md
wtrocki: lets leave it

wtrocki (approved):

PR #36, opened 2018-04-25
Adds refined JS SDK diagram

josemigallas:


StephenCoady (approved):
excellent diagram, really shows it is extensible service-wise.

paolobueno (changes requested):
Awesome stuff, better way to distinguish between sdk code and existing modules.

Can you just update the html file alongside with the svg?

josemigallas:
Done @paolobueno thanks!

paolobueno (approved):

wtrocki:
Late for the party but:

PR #40, opened 2018-06-24
[WIP] Data Sync Proposal

david-martin:


sync/server.md
```diff
@@ -0,0 +1,403 @@
+# Sync Server
```
david-martin, line 1: ping @aliok @darahayes for thoughts on the proposal so far

sync/server.md
```diff
@@ -0,0 +1,403 @@
...
+})
+```
+
+As the server needs to store resources, such as the Data Schema, Resolver Mappings and Webhooks, a data store is required.
+The Data Sync server is intended to run in a Kubernetes environment such as Openshift.
+This allows the Data Sync server to use a ServiceAccount token to save these resources in 1 or more Secrets.
+This will avoid the need for a potentially heavyweight database, at least until something different is required.
+
+A pubsub service will be required for managing the event driven nature of Subscriptions when the Data Sync server is scaled beyond 1 Replica.
+A reasonable candidate for this is Redis and its set of [pubsub commands](https://redis.io/topics/pubsub).
```
david-martin: This is still TBD.
One approach I'm thinking of is using postgres and https://github.com/voxpelli/node-pg-pubsub.
Postgres could then be the store for the Data Schema, Resolver Mappings, Data Sources, Webhooks etc..., *and* be the pubsub mechanism

sync/server.md
```diff
@@ -0,0 +1,403 @@
...
+app.use('/graphql', bodyParser.json(), function (req, res, next) {
+  const schema = require('./lib/schemaParser').parseFromFile(SCHEMA_FILE, DATA_SOURCES_FILE, RESOLVER_MAPPINGS_FILE)
+  const graphql = graphqlExpress({ schema, tracing })
+  graphql(req, res, next)
+})
+```
+
+As the server needs to store resources, such as the Data Schema, Resolver Mappings and Webhooks, a data store is required.
+The Data Sync server is intended to run in a Kubernetes environment such as Openshift.
+This allows the Data Sync server to use a ServiceAccount token to save these resources in 1 or more Secrets.
```
david-martin: After some discussion with @darahayes, this is unlikely to be the way this works.
Up for discussion, but at the moment I'm thinking of shipping postgres with the data sync server. See comment below for more context https://github.com/aerogear/proposals/pull/40/files#r198116098

sync/sdk.md
```diff
@@ -0,0 +1,81 @@
+# Sync SDK
```
david-martin, line 1: @danielpassos @wtrocki I'd appreciate and feedback or input into this early draft of the sync SDK and how the memeolist app could use it.
See @secondsun proposal for more info https://github.com/aerogear/proposals/pull/41

sync/server.md
```diff
@@ -0,0 +1,402 @@
...
+
+More information about the GraphQL terminology and schema can be found at https://graphql.org/learn/schema/
+
+## Server Technology
+
+The Sync Server will provide a GraphQL API for Apps to store and retrieve data from, based on a defined Data Schema.
+The server will be written in Node.js, leveraging various existing graphql libraries, such as `graphql`, `graphql-subscriptions`, `graphql-tools`, `apollo-server-express` and `subscriptions-transport-ws`.
+The server will use the standard module mechanism, `npm`, for installing and managing these packages.
+Npm `scripts` will be used extensively for running CI and local development tasks, executing cli tools directly e.g. `nodemon`, `standard`, `mocha`.
+Build tools such as `grunt` or `gulp` will *not* be used as they introduce extra overhead where a simple cli command will suffice. See https://www.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/ for more info.
```
darahayes: +1000

sync/server.md
```diff
@@ -0,0 +1,404 @@
...
+**Resolver Mapping** - The definition for how to load/save data to a Data Source based on parameters sent by a Client, and how to format it before returning it to the client. All mapping defintions are compiled and executed as a Handlebars template. This allows for some logic when defined the request & repsonse mapping e.g. using the `toJson` help to stringify the result of an SQL query.
+
+**Resolver Request Mapping** - The defintion for how to load/save data to a Data Source based on parameters sent by a Client e.g. a SQL query, a HTTP endpoint, method & body.
+
+**Resolver Response Mapping** - The definition for how to format data retrieved from a Data Source before returning it to the client e.g. simply stringify an array of objects.
+
+**Connector** - The combination of a Data Source and Resolver Mapping capabilities e.g. a Postgres Connector would allow a developer to connect to a Postgres database, and define SQL queries to execute when particular GraphQL Queries or Mutations are called.
+
+More information about the GraphQL terminology and schema can be found at https://graphql.org/learn/schema/
+
```
aliok, line 22: What's a "schema"?  https://github.com/aerogear/aerogear-data-sync-server/blob/0a6b9941e28fe3ff3be7d815ab96629e7befeacc/examples/schema.example.graphql

david-martin: A GraphQL Schema is the defintion/interface for all Types in your system, and the interactions that can be made on them.

This blog post is quite good at explaining https://www.prisma.io/blog/graphql-server-basics-the-schema-ac5e2950214e/

The schema you linked to is a very simple example. 
There is 1 entity/model, `Note`, with 4 fields, `id`, `title`, `content`, `timestamp`.
You can list & read Notes (Queries)
and you can create, update & delete Notes (Mutations)

GraphQL Resolvers have the actual implementation for a given Query or Mutation.

sync/server.md
```diff
@@ -0,0 +1,404 @@
...
+* Configure Resolvers for any Queries or Mutations
+* Update/modify the Data Schema as needed, continuously saving and trying out the Schema
+
+The last part of the flow is very important.
+This implies a fast feedback loop when developing your Data Schema.
+To ensure the feedback loop is as fast as possible, the architecture of sync will be slightly different that the proof of concept architecture.
+Both the backend for the Admin UI, and the GraphQL API that Apps talk to will be part of the 1 Data Sync server.
+This server will continously run, without any restart needed when the Data Schema changes.
+To facilitate this, the mounting of the graphql express middleware will need to be done in such a way that it always rebuilds the schema based on the latests available. In reality, the parsed schema should be cached, and only reloaded/reparsed if it has changed. Here is a sample graphql endpoint mounting in express:
+
```
aliok: What's the idea here to make sure schema changes are picked up? 
Do we want to have an endpoint like "/updateSchema"? Or, something like store Data Schema, Resolver Mappings, Data Sources, Webhooks in configs/secrets and watch them using K8s mechanisms (like we do at push operator: https://github.com/aerogear/ups-config-operator/blob/a0be10d159752ad78a3897bd0645dfb7fa1d7502/pkg/configOperator/kubeHelper.go#L44)

sync/server.md
darahayes: I've also done a little bit of searching here. It looks like MongoDB also has some pubsub capabilities using a feature called tailable cursors. The following blogpost has a high level overview with examples written in Node.

http://tugdualgrall.blogspot.com/2015/01/how-to-create-pubsub-application-with.html

Just throwing it out there as an option.

sync/server.md
david-martin: > What's the idea here to make sure schema changes are picked up?

Correct.

> Do we want to have an endpoint like "/updateSchema"? Or, something like store Data Schema, Resolver Mappings, Data Sources, Webhooks 

Yes

> in configs/secrets and watch them using K8s mechanisms (like we do at push operator: https://github.com/aerogear/ups-config-operator/blob/a0be10d159752ad78a3897bd0645dfb7fa1d7502/pkg/configOperator/kubeHelper.go#L44)

Probably not this bit. See comment https://github.com/aerogear/proposals/pull/40/files#r198116375

sync/server.md
```diff
@@ -0,0 +1,404 @@
...
+The defintion for a PostgreSQL Data Source will have the following parameters:
+
+* Database Server Host
+* Database Server Port
+* Database Name
+* Database Username
+* Database Password
+
+Table names will not be configured as part of the Data Source for Postgres.
+Instead, the table will be determined by the SQL query defined in the Resolver Request Mapping.
```
aliok, line 104: Agreed, that's a great idea.
We have this in POC schema already:
```
type Note {
  id: String
  title: String
  content: String
  timestamp: Float
}
```
We would just make the id field a GraphQL id.

darahayes: Quick question about this though - How do we ensure the table is created. Let's say the user does the following:

* defines a schema
* defines a postgres data source (no table config)
* defines the a request mapping to `SELECT * from foo`

How do we ensure that the `foo` table gets created in the first place?

sync/server.md
```diff
@@ -0,0 +1,404 @@
...
+
+The above is not intended to limit what metrics to capture, but more as a starting point to discover what metrics are valueable.
+
+### Client Metrics
+
+There are 2 potential approaches to getting Client Metrics from a device to the Metrics service. The first approach is to use the MetricsService module in the AeroGear SDK to send metrics directly to the Metrics service. This would mean the Sync module in the SDK would need to call the Metrics service every so often.
+This is not a great approach as the device may not be online all the time.
+
+The second approach is to piggyback on any GraphQL requests, adding metrics data to the request context. The data will need to be limited in size to avoid unnecessary network overhead. Using simple counters that are maintained on the Client will help keep the payload size small. For example, the Sync module in the SDK maintains a specific set of counters as the App interacts with it.
+Whenever a Sync request (query/mutation) is made to the server, the latest values for this set of counters is sent along.
```
wei-lee, line 319: Is this similar to  the "__fh" thing we did in the old fh js sdk? If so I don't think it is a good idea. If the whole request is defined by the developers, then I don't think we should change it while it's being sent from the clients.

I would prefer to reuse the existing metrics module and service, and look at what changes we need to do to improve it - it will also give use the opportunity to improve the metrics module to make it more generic and useful for mobile developers.

sync/server.md
aliok: I meant, maybe document that as well in the terminology :)

sync/server.md
aliok: +1

sync/server.md
aliok: What I understand from the comment https://github.com/aerogear/proposals/pull/40/files#r198175673 , we will have a special endpoint which the schema is posted. I think that should be the step to create tables.

BTW, we also need to alter the tables if types are changed in the schema. ...which should be easy.

sync/server.md
```diff
@@ -0,0 +1,404 @@
...
+If you need to iterate through the response as a JSON array, mapping fields, you can do something like below.
+In this example, each item has a number of fields, one of which is `name`.
+However, this needs to be changed to `firstname` to match our schema.
+
+```
+TODO: Helper for mapping an array
+TODO: Is this a reasonable use case for response mapping? Maybe filtering is better e.g. omitting records that don't have a paritcular field set
+
+{{toJSON mappedResult}}
+```
```
wei-lee, line 222: @david-martin I think the default request and response resolvers are great for certain use cases, but I am not sure they are flexible enough to support some real-world use cases.

For example, in my resolver, I want to first fetch data from an external source via http. But the external source could be unstable, so I would like to cache/persist the data locally and if the external source is not available, the cached version of the data will be returned.

Is it possible to support this scenario in the resolvers?

I am also thinking can we allow developers to actually define the resolvers in nodejs? We can take advantage of the [VM module](https://nodejs.org/api/vm.html) to run the scripts in a sandboxed environment, but it will allow developers to do more if their use case is a bit more complicated. 

Do you think it is something worth spiking?

sync/server.md
```diff
@@ -0,0 +1,404 @@
...
+
+```
+{
+  "path": "/notes?id={{context.arguments.id}}",
+  "method": "GET",
+  "headers": {
+    "accept": "application/json"
+  }
+}
+```
```
wei-lee, line 197: is it possible to use websockets in this http resolver?

david-martin: I would see websockets being a separate data source

sync/server.md
david-martin: For custom code, there are 2 options:

* an openwhisk data source with custom code defined as actions
* write a custom server that exposes a http endpoint, and use the http data source

sync/server.md
david-martin: After off-thread discussions, using the metrics module in the client SDK makes sense if it is improved to add 2 features:

* custom metrics
* offline caching of metrics

sync/server.md
david-martin: @aliok data sync is not concerned with the data source, beyond configuring how to connect to it.
Data sync will have supported data sources e.g. postgres & HTTP,
but only in the context of resolver mappings.
Data Sync will not ship with a db backend for storing app/user data.

sync/server.md
david-martin: @johnfriz this is what i'd like to discuss with you in person tomorrow.
For subscriptions in a horizontally scaled environment where Clients are connected to different instances of the server, we need a pub/sub mechanism for event notification between the server instances.

Approaches discussed so far:

* zeromq for in-memory queues, but this gets hairy when you scale up to many server instances as each instance has to connect to every other instance
* redis pub/sub https://redis.io/topics/pubsub
* AMQ
* Postgres NOTIFY/LISTEN https://github.com/voxpelli/node-pg-pubsub, https://www.postgresql.org/docs/9.2/static/sql-notify.html, https://www.postgresql.org/docs/9.2/static/sql-listen.html
* etcd (watch)
* rethinkdb

Separately, there were also some discussions about where to store the data schema, resolver mappings, webhooks etc...
In the early architecture these were going to be stored in a Secret in openshift, then mounted into the graphql server.
However, I'm questioning this approach now that the architecture is simplified (see diagram in PR) with hot-reload of the schema as its changed (no redeployment necessary)
It puts a hard dependency of k8s/openshift on the server part, which wasn't the case before.
A few things that were discussed for this:

* just store them in a secret if the server is running in OpenShift. If it's not, load/save to a file on disk (e.g. for easy local dev)
* store them in a db
  * etcd (may also be used to provide pubsub)
  * postgres (may also be used to provide pubsub)
  * rethinkdb (may also be used to provide pubsub)

sync/server.md
wei-lee: so how does subscription work for http resolvers?

david-martin: If you take a look at the example in this readme
https://github.com/apollographql/graphql-subscriptions#getting-started-with-your-first-subscription

and another example in this simple server
https://github.com/david-martin/starwars-server/commit/af5804d177700b6e1fd0b79fe2bc3967cd5f6861

In GraphQL, a subscription can notify clients whenever another client has executed a query or mutation.
The subscription is not at the database/data source level, it's within the graphql server as requests pass through it.

So regardless of the data source (http, postgres, openwhisk, in-memory db), this is the flow:

* client sends a Mutation (createNote) to create a Note object
* the server executes the appropriate resolver
* in that resolver, an event is published that this mutation was executed (successfully)
* separately another client connects to the server over a websocket, and is subscribed to a Subscription (noteCreated)
* in that subscription resolver on the server, it is subscribed to any 'createNote' events
* as it hears about any 'createNote' event, it notifes any connected clients that are subscribe to this, and can include some payload (depending on what the schema defines)

More info here https://www.apollographql.com/docs/apollo-server/v2/features/subscriptions.html

sync/server.md
```diff
@@ -0,0 +1,404 @@
...
+
+String version
+```
+{\n  \"path\": \"/notes\",\n  \"method\": \"GET\",\n  \"headers\": {\n    \"accept\": \"application/json\"\n  }}
+```
+
+By treating the defintion as a string, it allows for Handlebars syntax to be added anywhere in the JSON object.
```
david-martin, line 186: @wei-lee you expressed some concern about the limitations of handlebars templates.
If we define exactly what we want developers to do in a request/response mapping, we should be able to choose the most appropriate template engine.
This site is handy for narrowing down what engine to use
http://garann.github.io/template-chooser/

My thinking is we want to limit what's possible in request/response mappings to data manipulation e.g. iterating, filtering, mapping.
For more complex logic, let the develop do that outside of data sync in their own service (and use the http data source), or hook up to OpenWhisk actions for custom code.
I have security concerns if we bring in a more flexible template engine that allows js code execution.

cc @johnfriz for thoughts

sync/sdk.md
```diff
@@ -0,0 +1,100 @@
...
+2 Apps were considered for this:
+
+* Memeolist App (more info at https://github.com/aerogear/proposals/pull/41)
+* Cool Stuff Store App
+
+At this time, only the Memeolist App will be used for this 'dogfooding' purpose. However, it may make sense to convert/write other Apps to work with Data Sync. 
+
+## Binary upload
+
+Binary upload in the context of GraphQL means supporting the upload of binary content from an App to a Mutation resolver.
```
danielpassos, line 25: Instead of send a _Blob_ why not convert it into a byte[]/String on the client side?

sync/server.md
wei-lee: @david-martin agreed, it should really depend what usecases we want to support in this version of sync. Maybe worth adding them to this proposal to make it clear?

sync/sdk.md
david-martin: Do you mean a base64 string?
While that will work, it sounds like a temporary solution/hack.
Ideally binary content (like an image) is transferred as binary to avoid unnecessary overhead, particularly if there's potential for network connectivity issues on mobile.

However, if the solution for supporting binary is too complex or not worth the effort, base64 may be the best solution

sync/server.md
david-martin: I'm thinking the following:

* can use template placeholders for arguments/results i.e. can access `context.arguments` & `context.result`
* can iterate through an array of results, filtering out entries based on some conditional (for authorisation)
* can iterate through an array of results, mapping entries from one structure to another (this may be more useful when dealing with non json data e.g. an xml doc
* can stringify json

sync/server.md
aliok: Ok, then what's the answer to Dara's question @david-martin ?

david-martin: > Quick question about this though - How do we ensure the table is created. Let's say the user does the following:
> 
> defines a schema
> defines a postgres data source (no table config)
> defines the a request mapping to SELECT * from foo
> How do we ensure that the foo table gets created in the first place?

We don't.
It is up to the developer to have their data source/database in the right condition for resolvers to work.
If the host is wrong, port is wrong, credentials are wrong, table is missing, fields are missing, data sync will report back an error, which the developer will have to fix.

sync/sdk.md
secondsun: We will probably want to separate the "sync" object from large binary objects.  This way we can quickly populate client side graphs without needing a lot of handling for updates failing because someone attached a 100MB zip file.  Of course with GraphQL this is easy because we can have a "nobinary" query that sync talks to and a "details" query that pulls down the large data of a single object.

For uploads, we want to avoid copying large data sets into RAM and uploading those, aside from that I don't have strong opinions of how things are uploaded.

sync/server.md
```diff
@@ -69,22 +69,25 @@ app.use('/graphql', bodyParser.json(), function (req, res, next) {
 })
 ```
 
-As the server needs to store resources, such as the Data Schema, Resolver Mappings and Webhooks, a data store is required.
-The Data Sync server is intended to run in a Kubernetes environment such as Openshift.
-This allows the Data Sync server to use a ServiceAccount token to save these resources in 1 or more Secrets.
-This will avoid the need for a potentially heavyweight database, at least until something different is required.
+The GraphQL server will be notified about changes via a `channel` in Postgres that it is `LISTEN`ing to.
+THe Admin server will send a `NOTIFY` to this channel whenever the shcema or resolver mappings have changed.
```
pb82: Typo: `shcema`

danielpassos:
@david-martin Is this still valid?

david-martin:
I doubt it as sync has probably moved on since this.

PR #44, opened 2018-07-10
Adding Coolstore proposal

trepel:
*Motivation*

This is first draft of proposal for using AeroGear on top of Coolstore demo. Thanks Wei & Evan for initial input. This serves for slightly different purposes that Summer's dogfood proposal - marketing and integration with existing (real-world) solution.

Comments/concerns/questions/objections welcome!

### DIFF
--- a/coolstore.md
+++ b/coolstore.md
@@ -1,0 +1,38 @@
+# Coolstore MSA with AeroGear
+[Coolstore MSA (MicroServices Architecture)](https://github.com/jbossdemocentral/coolstore-microservice) is a well established demo of using microservices implemented by various frameworks (not limited to Red Hat portfolio) running on OpenShift together to provide a complex solution, an e-shop in this particular case.
+
+AeroGear project might benefit from building on top of the Coolstore MSA to demonstrate the capabilities it provides. This might be very helpful due to following:
+
+- Coolstore MSA is a widely used demo for OpenShift and Middleware. Becoming part of it will promote AeroGear project to a broader audience both within and outside Red Hat.
+- Even though implemented by Red Hatters mostly it is the very real world scenario. It is desirable to know how much effort is required to integrate AeroGear capabilities into existing solution. AeroGear project is mostly being used for creating the backend from scratch at the moment, not much attention is paid to integration with existing solutions.
+- The upskilling is another benefit. Mobile applications for various platforms (Android, iOS, Xamarin, Cordova) can be created and it is a good opportunity for people upskilling in these areas by implementing them.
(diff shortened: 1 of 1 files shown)


coolstore.md
```diff
@@ -0,0 +1,38 @@
...
+- Implement mobile application using configuration provided by the Mobile Client and the appropriate AeroGear SDK
+
+The mobile application itself should be able to
+- Log in and log out, using Keycloak as the authentication broker
+- Browse the products
+- Display the product detail
+- Add products to cart
+- Check out an order
+
+It is probably better to start with Angular/Ionic mobile app since it is consumable by most people (North America is iOS mostly, while other geographical locations are mostly Android), and allows us to tell a Progressive Web App story via a single codebase - it also avoids iOS certificate issues. Android might be preferred for native since the initial setup is easier (also avoids certificate issues, etc.) and Java language is widely adopted within Red Hat.
```
wojta, line 28: I think that AeroGear JS SDK is not yet for pure-browser JavaScript, so we can't talk about Progressive Web App.

wojta: > Android might be preferred for native since the initial setup is easier (also avoids certificate issues, etc.) and Java language is widely adopted within Red Hat.
I think that for Coolstore we should go with Kotlin and maybe call it Koolstore :D And Java is not that cool as Kotlin is.

trepel: Thanks for feedback. I don't mind using Kotlin. I would leave such decisions on the person implementing the app. As long as they use the Aerogear SDK in recommended way.

trepel: @wojta Are you saying that you have to use Cordova in order to use Aerogear JS SDK ?

evanshortiss: > Are you saying that you have to use Cordova in order to use Aerogear JS SDK

Ah, after looking I believe that is the case yes, but for a PWA we can update it later once the JS SDK is updated. The application could still work as PWA, just would not use the SDK when running in web browser - would be possible use Angular providers to achieve this I think.

```
providers: [
    { provide: SecurityService, useClass: isBrowser ? SecurityStub : AeroGearSecurity }
] 
```

evanshortiss: TLDR - We can still create a progressive web application and native from single codebase meaning the demo is easier and we cover more bases.

danielpassos:
@wei-lee Do you think it still valid? Should we merge it?

wei-lee:
@danielpassos Coolstore is a bit out of date now. There is a new demo across M/W that is being developed atm and we probably will use that. So we can close this one.

PR #46, opened 2018-07-26
Updates to schema

wtrocki:
Target for this is to update schema to be minimal and sensible to be implemented on the server. Additionally I have removed things that were hard or not possible on the server. Structure of the data is not flexible for some functionalities but we do not target here to build fully functional app. Just build an sandbox for us to test most of the graphql corner cases.

We could extend that over the time to match the requirements.
It's best to use proposal to stage that as schema on the server were out of the sync

### DIFF
--- a/dogfood.md
+++ b/dogfood.md
@@ -64,62 +64,45 @@ The iOS SDK has all of the features we need, and iOS has an Apollo GraphQL libra
 
 The following is a graphql schema for the proposed data model.  The GraphQL schema also includes the public APIs that will be implemented.
 
-### Sample Schema
+### Current Schema
 
 ```graphql
 type Profile {
   id: ID! @isUnique
   email: String! @isUnique
-  display_name: String!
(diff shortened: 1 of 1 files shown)


wtrocki:
Working on implementing this on the server first.

danielpassos:
@wtrocki my only thought is if we put the like count into the Meme type when listing the memes there is no way to know if I already liked that meme.

wtrocki:
@danielpassos yeah, but this is going to evolve over the time. 

Just want to add more stuff as:
- graph.cool is not suitable for our usage
- server side has just single object with single field implemented.

aliok (approved):
Looks good @wtrocki 
I like that we're doing it progressively.

darahayes (approved):

danielpassos:
@wtrocki Gotcha, I'd say, let's not have like so. I think comments is enough to test the relationship scenario.

